PR #2, opened 2016-02-03
add pcrupto.go

vashstorm:


vashstorm:
add pcrypto and pcrypto_test

PR #4, opened 2016-02-26
fix bug

Hurricanezwf:
(1) add hearbeat echo from server to client


Hurricanezwf:
(1)add hearbeat echo from server to client

PR #6, opened 2016-02-29
enable client to connect non-local [ip:port]

Hurricanezwf:
enable client to connect non-local [ip:port]

### DIFF
--- a/src/frp/models/client/config.go
+++ b/src/frp/models/client/config.go
@@ -66,6 +66,11 @@ func LoadConf(confFile string) (err error) {
 				return fmt.Errorf("Parse ini file error: proxy [%s] no passwd found", proxyClient.Name)
 			}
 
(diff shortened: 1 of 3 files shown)


Hurricanezwf:
enable client to connect non-local [ip:port]

src/frp/models/client/config.go
```diff
@@ -66,6 +66,11 @@ func LoadConf(confFile string) (err error) {
 				return fmt.Errorf("Parse ini file error: proxy [%s] no passwd found", proxyClient.Name)
 			}
 
+			proxyClient.LocalIp, ok = section["local_ip"]
+			if !ok {
+				return fmt.Errorf("Parse ini file error: proxy [%s] no local_ip found", proxyClient.Name)
```
fatedier: If no ip set, use 127.0.0.1 as default.

PR #185, opened 2016-12-20
Update frps.ini

lxhao61:


fatedier:
`conf/frps.ini` is just a example configure file, you should push your changes to your own repo.

PR #230, opened 2017-01-13
Fix Bug for issure #227

LitleCarl:
Assure state is `Working` before releasing and deleting ProxyServer instance when `privilege_mode = true `in order that if there is another frpc client is online with same proxy name, it will not be deleted from `ProxyServers` map.


fatedier:
你这个怎么多了这么多 commit 记录，建议清理一下。

PR #311, opened 2017-05-06
Make it clear that HTTPS can be password-protected too

jeremejevs:
Had to [look in the code](https://github.com/fatedier/frp/blob/3f9749488a2c485ebcb6f10b4a76b04b679205cc/src/models/server/server.go#L186) to make sure that HTTPS can be password-protected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -376,7 +376,7 @@ Anyone who can guess your tunnel URL can access your local web server unless you
 
 This enforces HTTP Basic Auth on all requests with the username and password you specify in frpc's configure file.
 
-It can be only enabled when proxy type is http.
+It can be enabled only when proxy type is http/https.
 
 ```ini
 # frpc.ini
--- a/README_zh.md
+++ b/README_zh.md
@@ -382,7 +382,7 @@ host_header_rewrite = dev.yourdomain.com
 
 frp 支持通过 HTTP Basic Auth 来保护你的 web 服务，使用户需要通过用户名和密码才能访问到你的服务。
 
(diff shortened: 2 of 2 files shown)


fatedier:
I'm not sure if HTTPS can be password-protected, can you tell me more about it?

frps just forward everything it received except the hostname from https request. You don't need to put your HTTPS certificate file in your frps server, so frps can't decrypt message and judge if the Authorization header is correct.

jeremejevs:
Okay, sorry, I was exploring Ngrok alternatives and made a hasty conclusion about HTTPS auth support when going over frp's code. Decided to open a doc PR and forgot about it (didn't have time to try frp yet). Though [a password is passed to `VhostHttpsMuxer.Listen`](https://github.com/fatedier/frp/blob/3f9749488a2c485ebcb6f10b4a76b04b679205cc/src/models/server/server.go#L188) and then [added to a `Listener` instance](https://github.com/fatedier/frp/blob/3f9749488a2c485ebcb6f10b4a76b04b679205cc/src/utils/vhost/vhost.go#L69), `VhostHttpsMuxer` [doesn't actually have an auth-checking function](https://github.com/fatedier/frp/blob/3f9749488a2c485ebcb6f10b4a76b04b679205cc/src/utils/vhost/https.go#L51) (the third `nil`), so HTTPS currently can't be auth-protected. Closing this PR.

I didn't question my conclusion because this didn't seem like a strange feature to have, password-protected HTTPS. Didn't know at the time that I can't give frp my private SSL key + I could see it utilizing Let's Encrypt to acquire the certs on its own (would be a nice option). My fault for not being more thorough.

PR #572, opened 2017-12-22
set required `server` role for server endpoint

bcho:
Base on [models/config/proxy.go](https://github.com/fatedier/frp/blob/a384bf55802d3bb0a91194978a0090fed26a5c6e/models/config/proxy.go#L635-L639), server role should be set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,6 +231,7 @@ Configure frps same as above.
 
   [secret_ssh]
   type = stcp
+  role = server
   sk = abcdefg
   local_ip = 127.0.0.1
   local_port = 22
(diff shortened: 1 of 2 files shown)


fatedier:
The default value of role is `server` changeg in https://github.com/fatedier/frp/commit/d873df5ca8ec070cf4faa15c94dc80c85291d94f.
So it's better let users not care about this param.

bcho:
ok, but current documentation's settings is broken for releases before (and including) v0.14.1, I think it's better to clarify this change for old releases (at least for current master). If you think this is irrelevant, feel free to close this pr.

PR #595, opened 2018-01-08
Use sans-serif font in web

NemoAlex:
I think sans-serif fonts are more suitable to this page.
Set a font-family to body will do the work.


NemoAlex:
Sorry I mess up the indentation. I will redo it.

Issue #636, opened 2018-02-16
Feature request: dashboard_api for getting a client status by name


miwee:
At present there is `api/proxy/traffic/:name` but it does not have info whether client is online or offline. The only workaround is to use `/api/proxy/tcp` (or udp, http etc.)  and get full data for all clients and parse specific info from this combined data.

From monitoring purpose, it is useful to have an api similar to  `api/proxy/status/:name`, which simply tells the current status and last known online time.

thanks

miwee:
I have changed and tested this feature. I'm attaching the patch. 
I was trying to use github workflow, but I couldn't compile my forked repository. It was giving some error related to hardcoded paths github.com/fatedier
Hope this should be ok.
[0001-dashboard_api-for-getting-a-client-status-by-name.txt](https://github.com/fatedier/frp/files/1873139/0001-dashboard_api-for-getting-a-client-status-by-name.txt)

fatedier:
Could you send a pr to dev branch? I will review it conveniently.

fatedier:
Just change directory name `miwee` to `fatedier` to compile it.

PR #688, opened 2018-04-04
dashboard_api for getting a client status by name

miwee:
against issue https://github.com/fatedier/frp/issues/636

### DIFF
--- a/server/metric.go
+++ b/server/metric.go
@@ -263,6 +263,37 @@ func StatsGetProxiesByType(proxyType string) []*ProxyStats {
 	return res
 }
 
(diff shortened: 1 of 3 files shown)


server/metric.go
```diff
@@ -263,6 +263,37 @@ func StatsGetProxiesByType(proxyType string) []*ProxyStats {
 	return res
 }
 
+func StatsGetProxiesByTypeAndName(proxyType string, proxyName string) (res *ProxyStats) {
```
fatedier: It's not efficient to iterate all proxy stats, but ok now. We should optimize this later.

PR #698, opened 2018-04-09
add service mode for frp

flyingtime:
frpc use default config file:frpc.ini and frps use default config file:frps.ini


fatedier:
Startup params will be refacted in next version. This is not useful for us now.

flyingtime:
@fatedier ok

PR #875, opened 2018-08-03
websocket protocol

jettyu:
for issue: [希望frpc登陆支持websocket协议 #871](https://github.com/fatedier/frp/issues/871)


cmd/frpc/sub/websocket.go
```diff
@@ -0,0 +1,85 @@
...
+	websocketCmd.PersistentFlags().IntVarP(&localPort, "local_port", "l", 0, "local port")
+	websocketCmd.PersistentFlags().IntVarP(&remotePort, "remote_port", "r", 0, "remote port")
+	websocketCmd.PersistentFlags().BoolVarP(&useEncryption, "ue", "", false, "use encryption")
+	websocketCmd.PersistentFlags().BoolVarP(&useCompression, "uc", "", false, "use compression")
+
+	rootCmd.AddCommand(websocketCmd)
+}
+
+var websocketCmd = &cobra.Command{
+	Use:   "websocket",
```
fatedier: It's not a sub command. "websocket" is a option of protocol.

jettyu:
@fatedier Ok! But if not copy the vendor folder, the check will be failed!

models/consts/consts.go
```diff
@@ -23,10 +23,11 @@ var (
...
-	HttpsProxy string = "https"
-	StcpProxy  string = "stcp"
-	XtcpProxy  string = "xtcp"
+	TcpProxy       string = "tcp"
+	UdpProxy       string = "udp"
+	HttpProxy      string = "http"
+	HttpsProxy     string = "https"
+	StcpProxy      string = "stcp"
+	XtcpProxy      string = "xtcp"
+	WebsocketProxy string = "websocket"
```
fatedier: Nothing about websocket proxy, you should only change codes about `protocol` how frpc connect to frps.

models/config/proxy.go
```diff
@@ -825,6 +825,55 @@ func (cfg *XtcpProxyConf) CheckForSvr() (err error) {
 	return
 }
 
+// WEBSOCKET
+type WebsocketProxyConf struct {
```
fatedier: It's useless.

models/consts/consts.go
jettyu: If reuse the bind_port, the websocket is use http,  when the bind_port == vhost_http_port, how to differentiate the request conn is login or is for vhost_http ?

fatedier: Yes, it's a good question and i will try to do that after this PR.

jettyu:
@fatedier I have deal with reuse  bind_port for websocket, when vhost_http_port == bind_port, I judge by the host, the code is in the server/service.go

fatedier:
We'd better split it into different pull requests since there may be someting ignored or mistake in `muxer` package. So i just need to review code about `websocket` in this PR.

PR #882, opened 2018-08-08
Fix the problem of long connection for more than 30 seconds and disconnection of the server

235832289:
Fix the problem of long connection for more than 30 seconds and disconnection of the server


fatedier:
1. Send PR to dev branch, not master.
2. `git rebase -i xxxx` and `git push -f xxx` will work, no need to open another PR.

PR #886, opened 2018-08-10
Connect protocol support websocket

fatedier:


server/service.go
```diff
@@ -137,6 +139,13 @@ func NewService() (svr *Service, err error) {
 		log.Info("frps kcp listen on udp %s:%d", cfg.BindAddr, cfg.KcpBindPort)
 	}
 
+	// Listen for accepting connections from client using websocket protocol.
+	websocketPrefix := []byte("GET /%23frp")
```
jettyu: We can not judge the protocol is websocket just by the prefix. I test when the bind_port equal the v_http_port,  it transports failed

fatedier: `/#frp` or other uniq string is a special path prefix of frp. It's a simple and efficient way to detect protocol used for frp in case reading too many letters or facing other problems.

We don't need to care aboud the coincidence if others used the same path which i think it has low probility. In other words, we could let the prefix more unique and short, maybe using some special letters.

server/service.go
```diff
@@ -149,23 +158,23 @@ func NewService() (svr *Service, err error) {
...
+			l = svr.muxer.ListenHttp(1)
 		} else {
 			l, err = net.Listen("tcp", address)
 			if err != nil {
 				err = fmt.Errorf("Create vhost http listener error, %v", err)
 				return
 			}
+			go server.Serve(l)
 		}
-		go server.Serve(l)
```
jettyu: This can not be deleted. and should delete the line 168. Otherwise when the bind_port == v_http_port,  I don't find the http server.Serve! the result of test is as my excepted

server/service.go
jettyu: OK! I understand!

fatedier: Do you have any idea on the sepcial path? Such as `(~!)f` ?

jettyu: I think "#frp" is already unique if not test for it

PR #974, opened 2018-11-16
Fix setting basic auth password doc

acrogenesis:
Fix setting basic auth password doc
`http_passwd` doesn't work it should be `http_pwd`


fatedier:
I changed the target branch from master to dev.

PR #975, opened 2018-11-16
add multiple network interface support for socks5 plugin

doggeddog:
Similar with wget(--bind-address) or cURL(--interface),
 you can bind with a specific network interface.

config file like this: 
```
[plugin_socks5_bind_addr]
type = tcp
remote_port = 6007
plugin = socks5
plugin_user = abc
plugin_passwd = abc
plugin_bind_addr = 172.21.0.107
```

### DIFF
--- a/conf/frpc_full.ini
+++ b/conf/frpc_full.ini
@@ -168,6 +168,17 @@ plugin = socks5
 plugin_user = abc
 plugin_passwd = abc
 
+[plugin_socks5_bind_addr]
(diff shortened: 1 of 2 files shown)


fatedier:
It seems not a common requirement. We don't want to support more complicated features with plugin.
If someone has other requirements, launching another professional socks5 program is a good choice.

doggeddog:
That's all right.

PR #978, opened 2018-11-19
Added configuration example

Milokita:
an example for allow access from both frp.com & www.frp.com


fatedier:
It's not important enouth to put it in README. Maybe post a blog about it is a good choice.

Milokita:
OK

PR #984, opened 2018-11-27
chore: use scratch as docker base image to shrink image size

skiloop:
Docker images can be shrink to very small size with scratch for go applications.

https://blog.codeship.com/building-minimal-docker-containers-for-go-applications/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,20 @@
-FROM golang:1.10
+FROM golang:1.10 AS base
 
-COPY . /go/src/github.com/fatedier/frp
+ENV CGO_ENABLED=0 
+ENV GOOS=linux
+ENV GOARCH=amd64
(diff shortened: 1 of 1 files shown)


fatedier:
There is already a `Dockerfile_multiple_build` file in root dir and i may remove docker support in this project in future.

skiloop:
> There is already a `Dockerfile_multiple_build` file in root dir and i may remove docker support in this project in future.

Docker support is quit useful in many case, I think is not a good idea to remove it. 
By the way, scratch is much smaller than alpine image too, compressed image can be as small as 9m.

fatedier:
You can put it in other repos about how to use frp in docker.
This repo will only reserve codes since it's not good maintained here.

Jefffish09:
@fatedier sorry that I cannot find the file `Dockerfile_multiple_build` now in root dir of the repo.

PR #1044, opened 2019-01-15
client: fixup msg.Login, add missing Hostname field

442hz:


fatedier:
It's a reserved param and have no use.
We don't set it at present.

442hz:
OK, got it.

PR #1046, opened 2019-01-15
frpc: fixup ini config parse problem in sub command `status` and `reload`

442hz:
frpc: fixup ini config parse problem in sub command `status` and `reload`

problem:
when run `frpc -c /path-to/frpc.ini status` or `frpc -c /path-to/frpc.ini reload`, 
we'll get the error result:
> parse ini conf file error: invalid INI syntax on line 1: /path-to/frpc.ini

since `func parseClientCommonCfg(fileType int, content string) (err error)` 
only accept **file content**, not a **filename**.

we should call `config.GetRenderedConfFromFile` first to avoid the problem.
just like the code in cmd/frpc/sub/root.go:
```go
	content, err = config.GetRenderedConfFromFile(cfgFilePath)
	if err != nil {
		return
	}
	g.GlbClientCfg.CfgFile = cfgFilePath

	err = parseClientCommonCfg(CfgFileTypeIni, content)
```

### DIFF
--- a/cmd/frpc/sub/reload.go
+++ b/cmd/frpc/sub/reload.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/fatedier/frp/client"
 	"github.com/fatedier/frp/g"
+	"github.com/fatedier/frp/models/config"
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


fatedier:
Thanks for your fix. Could you make this change to dev branch?

442hz:
OK

Issue #1029, opened 2018-12-29
[Feature request] custom login_fail_exit interval


CzBiX:
**Describe the results you expected:**
Add option to custom sleep time after login failed.

**Can you point out what caused this issue (optional)**
10s is too frequently .

fatedier:
No plan for this and the feature may be removed in the futrue.
You can write a shell or use other manage program to implement the same behavior.

PR #1089, opened 2019-02-14
Changed LoginFailExit to LoginFailRetry

Mewster:
When client attempts a connection to the server, attempts a certain amount of times before exiting.

### DIFF
--- a/client/service.go
+++ b/client/service.go
@@ -75,16 +75,21 @@ func (svr *Service) GetController() *Control {
 
 func (svr *Service) Run() error {
 	// first login
+	retries := g.GlbClientCfg.LoginFailRetry
 	for {
 		conn, session, err := svr.login()
 		if err != nil {
 			log.Warn("login to server failed: %v", err)
 
-			// if login_fail_exit is true, just exit this program
+			// if retries exceeded login_fail_retry, just exit this program
 			// otherwise sleep a while and try again to connect to server
-			if g.GlbClientCfg.LoginFailExit {
+			if retries == 0 {
 				return err
 			} else {
+				//if retries < 0 we always retry
+				if retries > 0 {
+					retries--
+				}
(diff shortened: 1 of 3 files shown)


Mewster:
I'm preparing another pull request to add backward compatibility with login_fail_exit

fatedier:
See https://github.com/fatedier/frp/issues/1029.

It's a deprecated feature and will be removed in the future. Using scripts or other systemctl tools to do the same thing.

Mewster:
> See #1029.
> 
> It's a deprecated feature and will be removed in the future. Using scripts or other systemctl tools to do the same thing.

If this will be removed in the future, how can one configure a similar setup without using external scripts? Is there a reason why it's going to be deprecated?

fatedier:
It's nothing about frp. We should focus on the primary thing.

Do not let one program do all things.

Mewster:
> It's nothing about frp. We should focus on the primary thing.
> Do not let one program do all things.

I'm sorry, but I can't understand why the client shouldn't be configurable to manage specific login use-case out of the box without resorting to external system tools. I know it's not a primary functionality, but I wouldn't discard this feature either without a reason...

fatedier:
It's not a feature for one specified program.
Systemd and other similar programs may be the best choice. I don't want to add features like this to make this project complicated.

Provide a script to do things like this is another choice.

Mewster:
> It's not a feature for one specified program.
> Systemd and other similar programs may be the best choice. I don't want to add features like this to make this project complicated.
> 
> Provide a script to do things like this is another choice.

Can systemd or a script know that there have been a failed login, and setup a login-retry rule?

fatedier:
If login failed, it will exit with a none zero code.

Mewster:
> If login failed, it will exit with a none zero code.

Understood, thanks.

PR #1269, opened 2019-05-30
resp.Body must be closed after function return

jiajunhuang:
whether it's success or fail, otherwise it will cause memory leak
ref: https://golang.org/pkg/net/http/


jiajunhuang:
BTW, removed the trailing `else` clause

fatedier:
Submit PR to dev branch.

PR #1271, opened 2019-05-30
resp.Body must be closed after function return

jiajunhuang:
whether it's success or fail, otherwise it will cause memory leak
ref: https://golang.org/pkg/net/http/


jiajunhuang:
move from https://github.com/fatedier/frp/pull/1269

cmd/frpc/sub/status.go
```diff
@@ -78,76 +78,77 @@ func status() error {
...
-		res := &client.StatusResp{}
-		err = json.Unmarshal(body, &res)
-		if err != nil {
-			return fmt.Errorf("unmarshal http response error: %s", strings.TrimSpace(string(body)))
-		}
+	}
+	if resp.StatusCode != 200 {
+		return fmt.Errorf("admin api status code [%d]", resp.StatusCode)
+	}
+	defer resp.Body.Close()
```
fatedier: Move to line 82 ?

tests/util/util.go
```diff
@@ -28,51 +28,51 @@ func GetProxyStatus(statusAddr string, user string, passwd string, name string)
...
-		}
-		for _, s := range allStatus.Tcp {
-			if s.Name == name {
-				return &s, nil
-			}
+	}
+	if resp.StatusCode != 200 {
+		return status, fmt.Errorf("admin api status code [%d]", resp.StatusCode)
+	}
+	defer resp.Body.Close()
```
fatedier: Move to line 32

PR #1373, opened 2019-08-10
Update frps_full.ini

budsone:


fatedier:
Why change `disable` to `enable`?

Issue #1387, opened 2019-08-15
Provide a High-level Go API


velovix:
I have a use-case where I need to spin up reverse proxy servers automatically on-demand. This would be very easy to do with FRP if it exposed an API for starting servers in code, but near as I can tell FRP is not designed to be used as a library. I wonder if there is interest in providing this high-level API to other Go programs. If there is, this is something I would like to work on.

The following are the problems I've identified that prevent FRP from being usable as a library, and what I would do to fix them. There may be more issues that I've yet to identify, but I would be more than happy to dig deeper if there's any interest in making this happen.

# The Problem

FRP has very friendly-looking functions like `server.NewService()` that, at first glance, look like exactly the kind of high-level API that I am describing. However, there are a few problems:

- FRP uses a global configuration object. This makes sense for a command line tool, but makes less sense for a library. Specifically, it prevents creating more than one server with different configurations.
- `server.NewService()` and `client.NewService()` expect to have access to static resources for the monitoring panel. Again, this works for a tool but not for a library.
- Code is mostly documented inside the implementation instead of as top-level comments viewable on GoDoc.

# The Solution

I am only familiar at a surface level with FRP's code, so I suspect some of what I'm saying may be unreasonable or undesirable for reasons I'm not aware of. That said, here is how I would solve the above problems:

- Pass around configuration objects to functions instead of using a global configuration object. This would allow for more multiple servers to be run at once and I think would lead to a more natural API.
- Either...
    - Only load assets if the dashboard is enabled. This would be a good enough solution if we don't expect people who use FRP as a library to want the dashboard.
    - Allow users to provide their own implementation of `http.FileSystem` or some other interface that would override the use of `statik`.
- Add library documentation. This one is pretty straightforward :)

fatedier:
Thank you for your valuable advise.

I'm glad to see this changes but i need sometime to find the suitable way to do these things.

> Pass around configuration objects to functions instead of using a global configuration object. 

I'm not sure if it's the suitable way to just make the global configuration object a function parameter. Or we can split it into some small objects for different modules.

> Only load assets if the dashboard is enabled. This would be a good enough solution if we don't expect people who use FRP as a library to want the dashboard.

Yes, this is easy to modify.

> Allow users to provide their own implementation of http.FileSystem or some other interface that would override the use of statik.

Now it support statik and static file path configured by `assets_dir`.

> Add library documentation

It's a hard work while it really lack of much documentation and unit tests.

velovix:
[...]

PR #1391, opened 2019-08-16
Load assets for dashboard/admin panel on demand

velovix:
The client and server services now only attempt to load assets if the dashboard or admin panel are enabled. This change makes it possible to use FRP as a library without having to manage assets.

If a library user wants to start a server with the dashboard enabled, they will need to set the DashboardPort and AssetsDir fields of ServerCommonConf.

If a library use wants to start a client with the admin panel enabled, that isn't currently possible because there's no way to customize the admin panel's assets directory. I can send a PR for that as well if there's interest.

Related: #1387

### DIFF
--- a/client/service.go
+++ b/client/service.go
@@ -52,13 +52,6 @@ type Service struct {
 }
 
 func NewService(pxyCfgs map[string]config.ProxyConf, visitorCfgs map[string]config.VisitorConf) (svr *Service, err error) {
-	// Init assets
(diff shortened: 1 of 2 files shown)


fatedier:
It's ok but you should check out new branch from dev and send PR to dev branch while master branch is used to release new version.

velovix:
Sorry, I missed that in the Contributing section, despite it being in bold. I'll open a new PR.

PR #1396, opened 2019-08-19
Pass server configuration as an argument

velovix:
The ServerCommonConf is now passed around as an argument instead of being shared between components as a global variable. This allows for more natural interaction with server.Session as a library and allows for multiple servers to co-exist within the same process.

This on its own is not an ideal solution because a lot more information is being passed around than is necessary. Splitting up this configuration information may be done as a future step by @fatedier.

Related: #1387

### DIFF
--- a/cmd/frps/root.go
+++ b/cmd/frps/root.go
@@ -20,7 +20,6 @@ import (
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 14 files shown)


fatedier:
LGTM
Thanks for your nice work! 

I reviewed the code i write before. It's not elegant. I will refactor these code in the near future.

PR #1644, opened 2020-02-10
DetailedErrorsToClient - only send detailed error info if this is on

GuyLewin:


server/service.go
```diff
@@ -331,9 +335,13 @@ func (svr *Service) HandleListener(l net.Listener) {
 				case *msg.NewVisitorConn:
 					if err = svr.RegisterVisitorConn(conn, m); err != nil {
 						xl.Warn("register visitor conn error: %v", err)
+						errStr := "register visitor conn error"
```
fatedier: Wrap a new function for this?

```golang
func GenerateResponseErrorString(summary string, err error, detailed bool) string {
    if detailed {
        ...
    } else {
        ...
    }
}
```

fatedier:
It has conflicts with #1643.

server/service.go
GuyLewin: Good point, will do

GuyLewin:
@fatedier  - what do you think, should we pass `xl` into the function and it'll log the warning too? Or is it too much coupling and we should keep it as is?

fatedier:
@GuyLewin 
> what do you think, should we pass xl into the function and it'll log the warning too? Or is it too much coupling and we should keep it as is?

I think it's ok now.

PR #1661, opened 2020-02-17
[Feature] TCP HTTP Tunnel

GuyLewin:
Use HTTP CONNECT headers on a single port to receive connections to any proxy


fatedier:
Is `tcpmux` better than `tcphttptunnel` ? Consider extensibility, we can support other multiplexer besides HTTP Connect in future (even custom multiplexer implemented by Plugin).

For example:
```ini
# frps.ini
[common]
bind_port = 7000
tcpmux_httpconnect_port = 1337
```

```ini
# frpc.ini
[common]
server_addr = x.x.x.x
server_port = 7000

[proxy1]
type = tcpmux
multiplexer = httpconnect
custom_domains = test1

[proxy2]
type = tcpmux
multiplexer = httpconnect
custom_domains = test2
```

```go
// github.com/fatedier/frp/server/proxy/tcpmux.go
type TcpMuxProxy struct {
        ...
        httpconnectMuxer
        otherMuxer
}

func (pxy *TcpMuxProxy) Run() (remoteAddr string, err error) {
        switch pxy.cfg.Multiplexer {
        case "httpconnect":
        ...
        case "other":
        ...
        default:
        }

        ...
        pxy.startListenHandler(pxy, HandleUserTcpConnection)
}
```

utils/vhost/tcphttptunnel.go
```diff
@@ -0,0 +1,67 @@
...
+	"net/http"
+	"time"
+)
+
+type TcpHttpTunnelMuxer struct {
+	*VhostMuxer
+}
+
+func NewTcpHttpTunnelMuxer(listener net.Listener, timeout time.Duration) (*TcpHttpTunnelMuxer, error) {
+	mux, err := NewVhostMuxer(listener, getHostFromHttpConnect, nil, sendHttpOk, timeout)
```
fatedier: The fourth params in `NewVhostMuxer` is `hostRewriteFunc`, it's tricky but ok now. Maybe add TODO comments to refactoring later.

GuyLewin:
> Is `tcpmux` better than `tcphttptunnel` ? Consider extensibility, we can support other multiplexer besides HTTP Connect in future (even custom multiplexer implemented by Plugin).
> 
> For example:
> 
> ```ini
> # frps.ini
> [common]
> bind_port = 7000
> tcpmux_httpconnect_port = 1337
> ```
> 
> ```ini
> # frpc.ini
> [common]
> server_addr = x.x.x.x
> server_port = 7000
> 
> [proxy1]
> type = tcpmux
> multiplexer = httpconnect
> custom_domains = test1
> 
> [proxy2]
> type = tcpmux
> multiplexer = httpconnect
> custom_domains = test2
> ```
> 
> ```go
> // github.com/fatedier/frp/server/proxy/tcpmux.go
> type TcpMuxProxy struct {
>         ...
>         httpconnectMuxer
>         otherMuxer
> }
> 
> func (pxy *TcpMuxProxy) Run() (remoteAddr string, err error) {
>         switch pxy.cfg.Multiplexer {
>         case "httpconnect":
>         ...
>         case "other":
>         ...
>         default:
>         }
> 
>         ...
>         pxy.startListenHandler(pxy, HandleUserTcpConnection)
> }
> ```

@fatedier the concept sounds great actually. Just to be clear - giving `frpc` the ability to specify what multiplexer to use per TcpMux would require us to listen to all possible types of multiplexes on different ports with different routing managers, and adding a proxy to the right routing manager based on this selection. I think that might be OK, but it would require `frps` to listen to all the multiplex types it's expecting to get from `frpc`s. What do you think?

utils/vhost/tcphttptunnel.go
GuyLewin: Yeah it's kind of a stretch. I thought about adding another parameter, just a callback for successful connections, instead of re-using this parameter. What do you think?

fatedier: I agree.

fatedier: I agree.

fatedier:
> the concept sounds great actually. Just to be clear - giving frpc the ability to specify what multiplexer to use per TcpMux would require us to listen to all possible types of multiplexes on different ports with different routing managers, and adding a proxy to the right routing manager based on this selection. I think that might be OK, but it would require frps to listen to all the multiplex types it's expecting to get from frpcs. What do you think?

@GuyLewin How about listening on port dynamically? 

Example:
```ini
[common]
server_addr = x.x.x.x
server_port = 7000

[proxy1]
type = tcpmux
local_port = 8080
remote_port = 6000
multiplexer = httpconnect
custom_domains = test1

[proxy2]
type = tcpmux
local_port = 8081
remote_port = 6000
multiplexer = httpconnect
custom_domains = test2

[proxy3]
type = tcpmux
local_port = 8083
remote_port = 6001
multiplexer = httpconnect
custom_domains = test1
```
Port is configured by frpc. 

Connection to port 6000 with host test1 -> proxy1.
Connection to port 6000 with host test2 -> proxy2.
Connection to port 6001 with host test1 -> proxy3.

Assume the proxies register order is proxy1, proxy2 and proxy3.

proxy1: Creata a new httpconnect multiplexer for port 6000 and register host test1.
proxy2: httpconnect multiplexer for port 6000 exists, just register new host test2.
proxy3: Creata a new httpconnect multiplexer for port 6001 and register host test1.

After all proxies releated with one port multiplexer closed, remove it.

GuyLewin:
> > the concept sounds great actually. Just to be clear - giving frpc the ability to specify what multiplexer to use per TcpMux would require us to listen to all possible types of multiplexes on different ports with different routing managers, and adding a proxy to the right routing manager based on this selection. I think that might be OK, but it would require frps to listen to all the multiplex types it's expecting to get from frpcs. What do you think?
> 
> @GuyLewin How about listening on port dynamically?
> 
> Example:
> 
> ```ini
> [common]
> server_addr = x.x.x.x
> server_port = 7000
> 
> [proxy1]
> type = tcpmux
> local_port = 8080
> remote_port = 6000
> multiplexer = httpconnect
> custom_domains = test1
> 
> [proxy2]
> type = tcpmux
> local_port = 8081
> remote_port = 6000
> multiplexer = httpconnect
> custom_domains = test2
> 
> [proxy3]
> type = tcpmux
> local_port = 8083
> remote_port = 6001
> multiplexer = httpconnect
> custom_domains = test1
> ```
> 
> Port is configured by frpc.
> 
> Connection to port 6000 with host test1 -> proxy1.
> Connection to port 6000 with host test2 -> proxy2.
> Connection to port 6001 with host test1 -> proxy3.
> 
> Assume the proxies register order is proxy1, proxy2 and proxy3.
> 
> proxy1: Creata a new httpconnect multiplexer for port 6000 and register host test1.
> proxy2: httpconnect multiplexer for port 6000 exists, just register new host test2.
> proxy3: Creata a new httpconnect multiplexer for port 6001 and register host test1.
> 
> After all proxies releated with one port multiplexer closed, remove it.

The thing is I'm trying to build a solution with multiple `frpc` installed on multiple places unaware of one another, and they won't be "trusted" and most decisions will be made by `frps`. 
In the above scenario, I'll have to verify that the port match between different proxies and different `frpc`s, since I really want to mux all of them together.
For my purpose I need either:
* an ability to enforce all proxies are exactly as I expect (tcpmux, type=httpconnect, remote_port=<something>) 
OR
* configure most of this on the server side (`frps`) and just let the `frpc` describe it wants the tcpmux option.

How do you think we should solve this?

server/proxy/tcphttptunnel.go
```diff
@@ -0,0 +1,83 @@
...
+	xl := pxy.xl
+	routeConfig := &vhost.VhostRouteConfig{}
+
+	defer func() {
+		if err != nil {
+			pxy.Close()
+		}
+	}()
+	addrs := make([]string, 0)
+	for _, domain := range pxy.cfg.CustomDomains {
```
mpeterson: Inside this `for` you start listening in several ports. If the function succeeds all is good, but what happens if you `return` on an error? Shouldn't you clean up after you on all listeners that you started?

server/proxy/tcphttptunnel.go
```diff
@@ -0,0 +1,83 @@
...
+			return
+		}
+		xl.Info("tcp http tunnel server listen for host [%s]", routeConfig.Domain)
+		pxy.listeners = append(pxy.listeners, l)
+		addrs = append(addrs, util.CanonicalAddr(routeConfig.Domain, pxy.serverCfg.TcpHttpTunnelPort))
+	}
+
+	pxy.startListenHandler(pxy, HandleUserTcpConnection)
+	remoteAddr = strings.Join(addrs, ",")
+	return
```
mpeterson: IMHO this function is a little bit too big to use naked returns.

server/service.go
```diff
@@ -215,6 +215,23 @@ func NewService(cfg config.ServerCommonConf) (svr *Service, err error) {
...
+	// Create tcp http tunnel muxer.
+	if cfg.TcpHttpTunnelPort > 0 {
+		var l net.Listener
+		l, err = net.Listen("tcp", fmt.Sprintf("%s:%d", cfg.ProxyBindAddr, cfg.TcpHttpTunnelPort))
+		if err != nil {
+			err = fmt.Errorf("Create server listener error, %v", err)
+			return
+		}
+
+		svr.rc.VhostTcpMuxer, err = vhost.NewTcpHttpTunnelMuxer(l, 30*time.Second)
```
mpeterson: The timeout is a good candidate for a constant with a comment on why that arbitrary number was chosen.

fatedier:
> The thing is I'm trying to build a solution with multiple frpc installed on multiple places unaware of one another, and they won't be "trusted" and most decisions will be made by frps.
In the above scenario, I'll have to verify that the port match between different proxies and different frpcs, since I really want to mux all of them together.
For my purpose I need either:
 an ability to enforce all proxies are exactly as I expect (tcpmux, type=httpconnect, remote_port=)
OR
configure most of this on the server side (frps) and just let the frpc describe it wants the tcpmux option.
How do you think we should solve this?

@GuyLewin As i said in email to you. You can write the [Server Manage Plugin](https://github.com/fatedier/frp/blob/master/doc/server_plugin.md) to change `remote_port` to the same port.

However, configure `tcpmux_httpconnect_port` on frps is also a acceptable choice now. To consider code refactoring later if more users need the ability to specify the port binding for tcpmux. It's not complicated and compatible with the current solution.

client/proxy/proxy.go
```diff
@@ -141,6 +146,35 @@ func (pxy *TcpProxy) InWorkConn(conn net.Conn, m *msg.StartWorkConn) {
 		conn, []byte(pxy.clientCfg.Token), m)
 }
 
+// TCP HTTP Tunnel
+type TcpHttpTunnelProxy struct {
```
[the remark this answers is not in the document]
GuyLewin: I tried doing that but too much of the current code isn't like this. I'm not going to convert everything now (unless @fatedier thinks strongly otherwise), and I think the code should have a single code style.

client/proxy/proxy.go
```diff
@@ -141,6 +146,35 @@ func (pxy *TcpProxy) InWorkConn(conn net.Conn, m *msg.StartWorkConn) {
...
+}
+
+func (pxy *TcpHttpTunnelProxy) Close() {
+	if pxy.proxyPlugin != nil {
+		pxy.proxyPlugin.Close()
+	}
+}
+
+func (pxy *TcpHttpTunnelProxy) InWorkConn(conn net.Conn, m *msg.StartWorkConn) {
+	HandleTcpWorkConnection(pxy.ctx, &pxy.cfg.LocalSvrConf, pxy.proxyPlugin, &pxy.cfg.BaseProxyConf, pxy.limiter,
```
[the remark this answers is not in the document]
GuyLewin: mpeterson - I agree, but this def isn't the scope of this PR since I never even changed this func!

client/proxy/proxy.go
```diff
@@ -141,6 +146,35 @@ func (pxy *TcpProxy) InWorkConn(conn net.Conn, m *msg.StartWorkConn) {
...
+
+	cfg         *config.TcpHttpTunnelProxyConf
+	proxyPlugin plugin.Plugin
+}
+
+func (pxy *TcpHttpTunnelProxy) Run() (err error) {
+	if pxy.cfg.Plugin != "" {
+		pxy.proxyPlugin, err = plugin.Create(pxy.cfg.Plugin, pxy.cfg.PluginParams)
+		if err != nil {
+			return
```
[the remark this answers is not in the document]
GuyLewin: mpeterson - I agree, but this def isn't the scope of this PR since I never even changed this func!

server/proxy/tcphttptunnel.go
GuyLewin: If there's an error (`err != nil`) `pxy.Close()` will be called, which calls l.Close() on every listener in the array.

server/service.go
GuyLewin: I didn't put a comment, @fatedier feel free if you want to

GuyLewin:
@fatedier please take a look again 😄

client/proxy/proxy.go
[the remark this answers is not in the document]
fatedier: It can be done after this PR merged.

models/config/server_common.go
```diff
@@ -259,6 +266,17 @@ func UnmarshalServerConfFromIni(content string) (cfg ServerCommonConf, err error
...
+		if v, err = strconv.ParseInt(tmpStr, 10, 64); err != nil {
+			err = fmt.Errorf("Parse conf error: invalid tcpmux_httpconnect_port")
+			return
+		} else {
+			cfg.TcpMuxHttpConnectPort = int(v)
+		}
+	} else {
+		cfg.TcpMuxHttpConnectPort = 0
+	}
+
```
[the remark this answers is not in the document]
mpeterson: I'm not even sure that you actually need `cfg.TcpMuxHttpConnectPort = 0` at the beginning ...

utils/vhost/vhost.go
```diff
@@ -29,22 +29,25 @@ import (
...
 	timeout        time.Duration
 	vhostFunc      muxFunc
 	authFunc       httpAuthFunc
+	successFunc    successFunc
 	rewriteFunc    hostRewriteFunc
 	registryRouter *VhostRouters
 }
 
-func NewVhostMuxer(listener net.Listener, vhostFunc muxFunc, authFunc httpAuthFunc, rewriteFunc hostRewriteFunc, timeout time.Duration) (mux *VhostMuxer, err error) {
+func NewVhostMuxer(listener net.Listener, vhostFunc muxFunc, authFunc httpAuthFunc, successFunc successFunc, rewriteFunc hostRewriteFunc, timeout time.Duration) (mux *VhostMuxer, err error) {
```
mpeterson: non-voting: all the callbacks should be moved into a struct and interfaces. 

I put this as non-voting because it's out of the scope of this patch, but perhaps you can add a `TODO` comment to get back to it later. A good example of what I mean is [the way kubernetes does it for informers](https://github.com/kubernetes/client-go/blob/eb0bad8167df60e402297b26e2cee1bddffde108/tools/cache/shared_informer.go#L138), in particular the [`ResourceHandler`](https://github.com/kubernetes/client-go/blob/eb0bad8167df60e402297b26e2cee1bddffde108/tools/cache/controller.go#L200:6)

models/config/server_common.go
```diff
@@ -59,6 +59,12 @@ type ServerCommonConf struct {
 	// requests. By default, this value is 0.
 	VhostHttpsPort int `json:"vhost_https_port"`
 
+	// TcpMuxHttpConnectPort specifies the port that the server listens for TCP
+	// HTTP CONNECT requests. If the value is 0, the server will not multiplex TCP
+	// requests on one single port. If it's not - it will listen on this value for
+	// HTTP CONNECT requests. By default, this value is 0.
+	TcpMuxHttpConnectPort int `json:"tcpmux_httpconnect_port"`
```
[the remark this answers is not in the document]
GuyLewin: You're right, but again - this will cause inconsistency with the current code using `int` for all port numbers (and for parsing them).
I suggest we should do that after merging in a different cleanup PR.

GuyLewin:
Thanks! Still don't have permissions to merge :)

PR #1666, opened 2020-02-19
[Feature] OIDC Authentication

GuyLewin:
Perform OIDC authentication (configurable to be instead of current token authentication) between frpc and frps. Authentication can be performed on login message and ping messages.
This is implemented by a new auth plugin system.


models/auth/oidc.go
```diff
@@ -0,0 +1,118 @@
...
+
+	return &OidcAuthProvider{
+		tokenGenerator:         tokenGenerator,
+		authenticateHeartBeats: authenticateHeartBeats,
+	}
+}
+
+func (auth *OidcAuthProvider) SetLogin(loginMsg *msg.Login) (err error) {
+	tokenObj, err := auth.tokenGenerator.Token(context.Background())
+	if tokenObj == nil {
```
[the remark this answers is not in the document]
GuyLewin: I'm afraid the library might change, having err == null and token == null scenario. What I actually care about is the token, how can I rely on having a token if I'm checking the `err`?

models/auth/oidc.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+	return &OidcAuthConsumer{
+		verifier:               provider.Verifier(&verifierConf),
+		authenticateHeartBeats: authenticateHeartBeats,
+	}
+}
+
+func (auth *OidcAuthConsumer) VerifyLogin(loginMsg *msg.Login) (err error) {
+	token, err := auth.verifier.Verify(context.Background(), loginMsg.PrivilegeKey)
+	if token != nil {
```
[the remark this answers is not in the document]
GuyLewin: I'm afraid the library might change, having err == null and token == null scenario. What I actually care about is the token, how can I rely on having a token if I'm checking the `err`?

models/auth/oidc.go
```diff
@@ -0,0 +1,118 @@
...
+		auth.subjectFromLogin = token.Subject
+		return
+	}
+	return fmt.Errorf("invalid OIDC token in login: %v", err)
+}
+
+func (auth *OidcAuthConsumer) VerifyPing(pingMsg *msg.Ping) (err error) {
+	if !auth.authenticateHeartBeats {
+		// if heartbeat authentication is disabled - don't verify
```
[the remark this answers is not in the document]
GuyLewin: Removed

models/auth/oidc.go
```diff
@@ -0,0 +1,118 @@
...
+	if tokenObj == nil {
+		return fmt.Errorf("couldn't generate OIDC token for login: %s", err)
+	}
+	loginMsg.PrivilegeKey = tokenObj.AccessToken
+	return
+}
+
+func (auth *OidcAuthProvider) SetPing(pingMsg *msg.Ping) (err error) {
+	if !auth.authenticateHeartBeats {
+		// if heartbeat authentication is disabled - don't set
```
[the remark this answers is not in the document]
GuyLewin: Removed

models/auth/oidc.go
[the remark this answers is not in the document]
GuyLewin: Nvm, I've read that it's best practice

models/auth/oidc.go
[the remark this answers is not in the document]
GuyLewin: Nvm, I've read that it's best practice

models/auth/oidc.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+}
+
+func (auth *OidcAuthProvider) SetLogin(loginMsg *msg.Login) (err error) {
+	tokenObj, err := auth.tokenGenerator.Token(context.Background())
+	if tokenObj == nil {
+		return fmt.Errorf("couldn't generate OIDC token for login: %s", err)
+	}
+	loginMsg.PrivilegeKey = tokenObj.AccessToken
+	return
```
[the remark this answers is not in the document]
GuyLewin: Done

models/auth/oidc.go
```diff
@@ -0,0 +1,118 @@
...
+		// if heartbeat authentication is disabled - don't set
+		return nil
+	}
+
+	tokenObj, err := auth.tokenGenerator.Token(context.Background())
+	if tokenObj == nil {
+		return fmt.Errorf("couldn't generate OIDC token for ping: %s", err)
+	}
+	pingMsg.PrivilegeKey = tokenObj.AccessToken
+	return
```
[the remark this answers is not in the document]
GuyLewin: Exported to a common function `generateAccessToken()`

models/config/client_common.go
```diff
@@ -60,6 +60,34 @@ type ClientCommonConf struct {
...
+	// authenticate frpc with frps. If "token" is specified - token will be
+	// read into login message. If "oidc" is specified - OIDC (Open ID Connect)
+	// token will be issued using OIDC settings. By default, this value is "token".
+	AuthenticationMethod string `json:"authentication_method"`
+	// AuthenticateHeartBeats specifies whether to include authentication token in
+	// heartbeats sent to frps. By default, this value is false.
+	AuthenticateHeartBeats bool `json:"authenticate_heartbeats"`
+	// AuthenticateNewWorkConns specifies whether to include authentication token in
+	// new work connections sent to frps. By default, this value is false.
+	AuthenticateNewWorkConns bool `json:"authenticate_new_work_conns"`
```
mpeterson: This was introduced in a recent patch set, perhaps add this to the README too?

server/service.go
```diff
@@ -431,6 +435,11 @@ func (svr *Service) RegisterWorkConn(workConn net.Conn, newMsg *msg.NewWorkConn)
 		xl.Warn("No client control found for run id [%s]", newMsg.RunId)
 		return
 	}
+	// Check auth.
+	if err := svr.authVerifier.VerifyNewWorkConn(newMsg); err != nil {
```
fatedier: The messages to start a new work connection between frpc and frps is:
```
frps   ReqWorkConn ->   frpc
       <- NewWorkConn
      StartWorkConn ->
```

There are no response for `NewWorkConn` to tell frpc that auth failed. Should we send `NewWorkConnResp` if auth failed or add error info in `StartWorkConn`? And then close the work connection here.

models/config/client_common.go
GuyLewin: True. done

models/auth/auth.go
```diff
@@ -0,0 +1,82 @@
...
+
+package auth
+
+import (
+	"github.com/fatedier/frp/models/config"
+	"github.com/fatedier/frp/models/consts"
+	"github.com/fatedier/frp/models/msg"
+)
+
+type BaseAuth struct {
```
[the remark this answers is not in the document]
GuyLewin: True. Done

server/service.go
GuyLewin: @fatedier We still need to close the connection on server side too. So we can respond with a response including an error and then close. The same goes for `pong` messages

GuyLewin:
@fatedier please take a look again when you get the chance

client/control.go
```diff
@@ -292,7 +298,9 @@ func (ctl *Control) msgHandler() {
 		case <-hbSend.C:
 			// send heartbeat to server
 			xl.Debug("send heartbeat to server")
-			pingMsg := &msg.Ping{}
+			pingMsg := &msg.Ping{
+				Timestamp: time.Now().Unix(),
```
fatedier: No need to set this if `AuthenticateHeartBeats` is false. Move it to `SetPing`?

server/service.go
```diff
@@ -438,6 +438,9 @@ func (svr *Service) RegisterWorkConn(workConn net.Conn, newMsg *msg.NewWorkConn)
 	// Check auth.
 	if err := svr.authVerifier.VerifyNewWorkConn(newMsg); err != nil {
 		xl.Warn("Invalid authentication in NewWorkConn message on run id [%s]", newMsg.RunId)
+		msg.WriteMsg(workConn, &msg.StartWorkConn{
```
fatedier: I maybe forgot to close workConn after an error occurred.

It seems better to return an error here and close workConn if err isn't nil.

```go
...
switch {
    case *msg.NewWorkConn:
        if err := svr.RegisterWorkConn(conn, m); err != nil {
            conn.Close()
        }
}
```

models/auth/auth.go
```diff
@@ -15,14 +15,103 @@
...
+
+func getDefaultBaseConf() baseConfig {
+	return baseConfig{
+		AuthenticationMethod:     "token",
+		AuthenticateHeartBeats:   false,
+		AuthenticateNewWorkConns: false,
+	}
+}
+
+func unmarshalBaseConfFromIni(conf ini.File) baseConfig {
```
fatedier: I prefer not put parse functions here. There is no need to coupled them together since how to unmarshal config is not releated to this package.

We may support more configure file types future. It will be a bit complicated to modify code everywhere.

client/control.go
GuyLewin: @fatedier - do you want the same for NewWorkConn?

models/auth/auth.go
GuyLewin: So you prefer the .ini unmarshal specifically should happen back in client_common and server_common?

fatedier: Yes.

client/control.go
fatedier: Yes.

client/control.go
```diff
@@ -134,7 +140,12 @@ func (ctl *Control) HandleReqWorkConn(inMsg *msg.ReqWorkConn) {
 	}
 
 	m := &msg.NewWorkConn{
-		RunId: ctl.runId,
+		RunId:     ctl.runId,
+		Timestamp: time.Now().Unix(),
```
mpeterson: (non-voting)

What is the purpose on having to handle `Timestamp` in this object and others? Why can't `GetAuthKey()` handle that itself internally? 

If the answer is that it can, then we should create another PR to modify that and simplify the code all over.

models/auth/auth.go
mpeterson: For common things, I agree. But for example, each option that is specific to an auth provider should be processed by that auth provider. So auth providers should also have an interface to parse the configuration and they should implement it.

server/control.go
```diff
@@ -454,9 +460,18 @@ func (ctl *Control) manager() {
...
+					ctl.sendCh <- &msg.Pong{
+						Error: "invalid authentication in ping",
+					}
+					return
+				}
 				ctl.lastPing = time.Now()
 				xl.Debug("receive heartbeat")
-				ctl.sendCh <- &msg.Pong{}
+				ctl.sendCh <- &msg.Pong{
+					Error: "",
```
mpeterson: it defaults to an empty string.

client/control.go
GuyLewin: +@fatedier: I'm kind of in favour of keeping Timestamp on any case. It's an informative time that's added to this request, unrelated to the fact it helps with auth.

server/control.go
GuyLewin: Yeah true, I thought it would be better explicitly but maybe it's a waste

models/auth/auth.go
GuyLewin: I agree with mpeterson, I think it's cleaner that `common` isn't aware of the OIDC / token specific configurations. And since we're doing that, I don't mind that the auth configuration resides here, like Proxy has unmarshalFromIni

fatedier: Not really. `AuthClientConfig` includes all auth provider's options. What one auth provider need to implement is how to validate options it needs, not how to parse it from configuration file. 

For example, once we want to support json format, we parse the `ServerConfig` once, no need to call each module's parsing funtions.

```go
type ServerConfig struct {
    ...
    Auth auth.AuthClientConfig `json:"auth"`
    ...
}

var config ServerConfig
json.Unmarshal(buf, &config)

authSetter := auth.NewAuthSetter(config.Auth)
```

client/control.go
GuyLewin: I wrote on NewWorkConn - I kind of think Timestamp is useful info in these communication packets. If you disagree I can move it like you suggested

models/auth/auth.go
GuyLewin: @fatedier but that will make server_common aware of every new protocol that we add, with every small configuration within, and that's the only reference this configuration will have outside of the `auth` module.

In addition, IDK if we want to give AuthClientConfig a name (e.g. Auth) because that suggests that the structure isn't flat, when in reality the .ini is flat.

client/control.go
fatedier: Some users feedback that they are sensitive to network traffic in embedded devices. I think we'd better remove it if not used.

models/auth/auth.go
mpeterson: Yeah, I think this doesn't scale up if new protocols will be added in the future. In a way by doing so packages are "leaking" implementation details.

server/control.go
mpeterson: In general I favor explicit over implicit too. In this case however, making the developer aware of the attribute Error here does not really add much. For me this is non-voting so as you wish :)

models/auth/auth.go
fatedier: Yes, you are right. It's annoying that we can't map the struct to ini file in a simple way.

Keep it here for ini now. I will try to support toml or other formats later.

GuyLewin:
Closed and re-opened because GitHub didn't update my recent push.

GuyLewin:
@fatedier thanks! I don’t think I have permissions to rebase or merge this since I don’t have write access to the repo. Can you do it?

models/auth/oidc.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+}
+
+func (auth *OidcAuthProvider) generateAccessToken() (accessToken string, err error) {
+	tokenObj, err := auth.tokenGenerator.Token(context.Background())
+	if err != nil {
+		return "", fmt.Errorf("couldn't generate OIDC token for login: %v", err)
+	}
+	return tokenObj.AccessToken, nil
+}
```
yuyulei, line 169: @GuyLewin I contact oauth2 and oidc recently, but meet some problem when I used oidc auth. 

From the code, frpc use `access_token` as `PrivilegeKey` when login operation, and then frps will verify the  `PrivilegeKey`. 

```
func (v *IDTokenVerifier) Verify(ctx context.Context, rawIDToken string)
```

My confusion: why use "IDToken" to verify "access_token". As far as I know, `access_token` is from Oauth2 provider's response, and `id_token` is from OIDC response.  It seems that `access_token`is uncorrelated to `id_token`. 

I maybe have a misunderstanding, but can you help me?

GuyLewin: Hey @yuyulei , sorry for the delayed response.
You're right that sometimes access_token isn't JWT, and that won't be supported by this library.
We needed to use `access_token` since we wanted to verify OIDC permissions as well to a specific resource, not only the identity of the login. Perhaps it's wrong to rely on access_token being JWT, what is the access_token type you're getting?

yuyulei: Thanks for your response, @GuyLewin . 

At first, I found some issues about oidc, so I tried AliyunCloud as third-party `oauth` provider,  but failed. It seems that the `access_token` of AliyunCloud is not generated by `JWT`. 

Frp'users may choose different `oauth` provider. We are not enable to guarantee these cloud providers use `JWT` to generate `access_token`. Usually in `oauth` protocal, We send `access_token` to providers, and they will help us to verify it. Is it right ?

So how about use `id_token` to verify authentication ? By `oidc` protocal, we get `access_token` and `id_token` from `oidc` providers, then we can decode `id_token` by `JWT`. I think this way will be more common, just as `oidc` protocal.

GuyLewin: @yuyulei as I mentioned in my previous comment, using id_token won't allow you to verify the user actually has permissions to access the FRP server. I think it'll only allow you to know who the user is

PR #1668, opened 2020-02-20
refactoring monitor code, support prometheus

fatedier:
Only care about first commit. Just remove vendor codes in second commit since we use go mod instead.


zyfdegh:
great job, nice code

server/service.go
```diff
@@ -251,8 +249,12 @@ func NewService(cfg config.ServerCommonConf) (svr *Service, err error) {
...
 		statsEnable = true
 	}
-
-	svr.statsCollector = stats.NewInternalCollector(statsEnable)
+	if statsEnable {
+		modelmetrics.EnableMem()
+		if cfg.EnablePrometheus {
+			modelmetrics.EnablePrometheus()
+		}
+	}
```
kingjcy: 这个不是两个又相互依赖了？考虑一下单独使用开启呢？

fatedier: 目前的配置参数不太合适，暂时先这么做了，影响不大。

PR #1714, opened 2020-03-21
Add audience parameter to token request for Auth0 support

marksteve:
Thank you for adding oidc as an authentication method! 

This is for making oidc authentication compatible with Auth0
https://auth0.com/docs/api-auth/tutorials/client-credentials

### DIFF
--- a/models/auth/oidc.go
+++ b/models/auth/oidc.go
@@ -17,6 +17,7 @@ package auth
 import (
 	"context"
 	"fmt"
+	"net/url"
 
 	"github.com/fatedier/frp/models/msg"
 
@@ -148,10 +149,11 @@ type OidcAuthProvider struct {
 
 func NewOidcAuthSetter(baseCfg baseConfig, cfg oidcClientConfig) *OidcAuthProvider {
 	tokenGenerator := &clientcredentials.Config{
-		ClientID:     cfg.OidcClientId,
-		ClientSecret: cfg.OidcClientSecret,
-		Scopes:       []string{cfg.OidcAudience},
-		TokenURL:     cfg.OidcTokenEndpointUrl,
+		ClientID:       cfg.OidcClientId,
+		ClientSecret:   cfg.OidcClientSecret,
+		Scopes:         []string{cfg.OidcAudience},
+		TokenURL:       cfg.OidcTokenEndpointUrl,
+		EndpointParams: url.Values{"audience": []string{cfg.OidcAudience}},
 	}
 
 	return &OidcAuthProvider{


fatedier:
You'd better check out dev branch and send PR to that branch.

@GuyLewin Would you have time to review this PR?

GuyLewin:
> You'd better check out dev branch and send PR to that branch.
> 
> 
> 
> @GuyLewin Would you have time to review this PR?

@fatedier sure!

marksteve:
@fatedier @GuyLewin rebased my branch and updated this pr to target dev

GuyLewin:
@fatedier @marksteve I've been reading more about this and turns out Auth0 took a non-standard approach and are using the `audience` parameter rather than the standard `scope` parameter (look here for another issue it creates with Postman testing: https://community.auth0.com/t/client-credentials-with-postman/34282 and with Swagger integration: https://github.com/swagger-api/swagger-ui/issues/5322).
The current code runs great on Azure Active Directory's OAuth 2.0 API, as well as Okta from what I've read.
If we want to support Auth0 we might consider a configuration parameter that only adds it when relevant, since it's not the standard approach. Even though it probably will never fail any token request - I don't know if we should always add this non-standard parameter just for Auth0.
@fatedier WDYT?

fatedier:
@GuyLewin I agree to add a new configuration parameter for this since it's non-standard.

marksteve:
@GuyLewin @fatedier agree on this! What should the config be? Was thinking of something like this:
```
oidc_provider = auth0
```

(noop for other providers, can add support for more providers)

fatedier:
> @GuyLewin @fatedier agree on this! What should the config be? Was thinking of something like this:
> 
> ```
> oidc_provider = auth0
> ```
> 
> (noop for other providers, can add support for more providers)

It's ok for me.

GuyLewin:
> > @GuyLewin @fatedier agree on this! What should the config be? Was thinking of something like this:
> > ```
> > oidc_provider = auth0
> > ```
> > 
> > 
> > (noop for other providers, can add support for more providers)
> 
> It's ok for me.

OK for me too, as long as any string is accepted here and an error won't be thrown if an unfamiliar oidc_provider is specified.

marksteve:
Sorry about the inactivity! Haven't really needed this feature after first playing with it. Can revisit if someone else requests for it.

fatedier:
OK. We can reopen this issue if someone else report the same problem.

JKounter:
i report the same problem, trying to use swaggerhub and Auth0, cannot authorize due to the audience issue.

chriscfox:
I also report the same problem.

PR #1757, opened 2020-04-15
[Feature] Load Balance TcpMux FRP Client Connections

GuyLewin:
Implement a simple load balancing strategy (random) to be able to hold multiple FRP client connections with the TcpMux plugin for the same domain and host.
It's only for TcpMux now becauase Http and Https are tightly coupled with the fact that there's only 1 connection per domain + location. We should break this coupling in the future.

### DIFF
--- a/server/service.go
+++ b/server/service.go
@@ -108,7 +108,7 @@ func NewService(cfg config.ServerCommonConf) (svr *Service, err error) {
 			TcpPortManager: ports.NewPortManager("tcp", cfg.ProxyBindAddr, cfg.AllowPorts),
 			UdpPortManager: ports.NewPortManager("udp", cfg.ProxyBindAddr, cfg.AllowPorts),
 		},
-		httpVhostRouter: vhost.NewVhostRouters(),
+		httpVhostRouter: vhost.NewVhostRouters(false),
 		authVerifier:    auth.NewAuthVerifier(cfg.AuthServerConfig),
 		tlsConfig:       generateTLSConfig(),
 		cfg:             cfg,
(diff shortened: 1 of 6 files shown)


GuyLewin:
@fatedier whenever you're free, this is a big issue for us. PS - I think `make ci` is failing on the `dev` branch.

fatedier:
Should we follow the definition of [Load Balancing](https://github.com/fatedier/frp#load-balancing) ? 

The main purpose to specify `group` and `group_key` is to prevent anyone can register the same domain and host in multiple users scenario .

GuyLewin:
@fatedier that's good! I wasn't aware of the `group` load balancing feature.
As you probably realized the use case I'm trying to solve is to load balance multiple `frcp`s using the same domain and location (actually they're all using our new `tcpmux`).
You think we should support `group` in all vhost load balancing implementations (http / https / tcpmux)?
Right now I didn't implement LB for http or https because it looked to complex to compare the structures between all definitions.

fatedier:
@GuyLewin I have already implemet load balancing for `http` in a different way since it doesn't expose a `Listener`. 

https://github.com/fatedier/frp/blob/master/server/group/http.go

`tcpmux` is different with `http`, but similar with `tcp` .

https://github.com/fatedier/frp/blob/master/server/group/tcp.go

We can implement a `TCPMuxGroupListener` to listen on specified host and redirect it to a random proxy(It's also easy to extend other algorithms). So we don't need to modify code in `vhost` package.

GuyLewin:
Look at #1765 for the `group` based implementation.

Issue #1761, opened 2020-04-16
StartWorkConn contains error: invalid authentication in NewWorkConn


jinkun2014:
Issue is only used for submiting bug report and documents typo. If there are same issues or answers can be found in documents, we will close it directly.

Use the commands below to provide key information from your environment:
You do NOT have to include this information if this is a FEATURE REQUEST

**What version of frp are you using (./frpc -v or ./frps -v)?**
0.32.0

**What operating system and processor architecture are you using (`go env`)?**
frpc in window 10 
frps in centos 8 

**Configures you used:**
frps.ini
```
[common]
bind_port = 7000
authentication_method = token
authenticate_heartbeats = true
authenticate_new_work_conns = true
token = 12345678
```
frpc.ini
```
[common]
server_addr = x.x.x.x
server_port = 7000
token = 12345678

[mall-server]
type = tcp
local_ip = 127.0.0.1
local_port = 9999
remote_port = 9999
```

**Steps to reproduce the issue:**
1. frps log
```
2020/04/16 11:06:43 [I] [service.go:157] frps tcp listen on 0.0.0.0:7000
2020/04/16 11:06:43 [I] [root.go:209] start frps success
2020/04/16 11:06:49 [I] [service.go:420] [102643e5bcabc23d] client login info: ip [61.144.96.178:65199] version [0.32.0] hostname [] os [windows] arch [amd64]
2020/04/16 11:06:49 [W] [service.go:462] Invalid authentication in NewWorkConn message on run id [102643e5bcabc23d]
2020/04/16 11:06:49 [I] [tcp.go:63] [102643e5bcabc23d] [mall-server] tcp proxy listen port [9999]
2020/04/16 11:06:49 [I] [control.go:444] [102643e5bcabc23d] new proxy [mall-server] success
```
2. frpc log
```
2020/04/16 11:06:48 [I] [service.go:257] [102643e5bcabc23d] login to server success, get run id [102643e5bcabc23d], server udp port [0]
2020/04/16 11:06:48 [I] [proxy_manager.go:144] [102643e5bcabc23d] proxy added: [mall-server]
2020/04/16 11:06:48 [E] [control.go:162] [102643e5bcabc23d] StartWorkConn contains error: invalid authentication in NewWorkConn
2020/04/16 11:06:48 [I] [control.go:179] [102643e5bcabc23d] [mall-server] start proxy success
```
3.

**Describe the results you received:**

Can not be accessed normally

**Describe the results you expected:**

[...]

PR #1762, opened 2020-04-16
fix: frps verify frpc NewWorkConn token failed 

bingtianbaihua:
https://github.com/fatedier/frp/issues/1761


utils/util/util.go
```diff
@@ -103,6 +103,9 @@ func ParseRangeNumbers(rangeStr string) (numbers []int64, err error) {
 }
 
 func GenerateResponseErrorString(summary string, err error, detailed bool) string {
+	if err == nil {
```
fatedier, line 106: Not required. Only call this function with an error.

puretrut:
0.33.0 win64 release 仍然存在这个问题呀

PR #1765, opened 2020-04-18
[Feature] Group TCPMux Proxies

GuyLewin:
Group TCPMux proxies together (currently only HTTP Connect multiplexer is supported) if configured with `group` and `groupKey`. This is useful to load balance multiple proxies with the same domain.


GuyLewin:
@fatedier this one should match the `group` architecture.

PR #1794, opened 2020-05-07
feat: Support user specify udp packet size in config

bingtianbaihua:
https://github.com/fatedier/frp/issues/1487

### DIFF
--- a/conf/frpc_full.ini
+++ b/conf/frpc_full.ini
@@ -68,6 +68,9 @@ tls_enable = true
 meta_var1 = 123
 meta_var2 = 234
 
+# specify udp packet size, unit is byte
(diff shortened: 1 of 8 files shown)


conf/frpc_full.ini
```diff
@@ -68,6 +68,9 @@ tls_enable = true
 meta_var1 = 123
 meta_var2 = 234
 
+# specify udp packet size, unit is byte
+udp_packet_size = 1024
```
fatedier, line 72: udp_packet_size = 1500

fatedier: More comments. 

This parameter should be same between client and server.

It affects the udp and sudp proxy.

PR #1798, opened 2020-05-09
fix: add frpc tls_enable flag and frps tls_only flag 

bingtianbaihua:
https://github.com/fatedier/frp/issues/1797


fatedier:
Add this for frps too?

cmd/frpc/sub/http.go
```diff
@@ -46,6 +46,7 @@ func init() {
 	httpCmd.PersistentFlags().StringVarP(&hostHeaderRewrite, "host_header_rewrite", "", "", "host header rewrite")
 	httpCmd.PersistentFlags().BoolVarP(&useEncryption, "ue", "", false, "use encryption")
 	httpCmd.PersistentFlags().BoolVarP(&useCompression, "uc", "", false, "use compression")
+	httpCmd.PersistentFlags().BoolVarP(&tlsEnable, "tls_enable", "", false, "enable frpc tls")
```
fatedier, line 49: You can improve this flag register section after this PR has been merged. 

It's not convenient to add new flag since we should add it to all different commands.

It's better to add a new function like `RegisterCommonFlags(cmd *cobra.Common)` and call it for each command.

bingtianbaihua: got it

PR #1855, opened 2020-06-15
Clients that support non-preemptive authentication

fzhyzamt:
example client: okhttp <= 3.11.0


fatedier:
1. You should checkout dev branch to develop.
2. Why we need this header?

fzhyzamt:
Non-preemptive authentication means that the Proxy-Authorization will not be actively sent when the tunnel is established, and this authentication will be brought after receiving the 407 response.

The server will now directly close the connection when the proxy verification fails, and the client will still send a request on this connection without knowing it, causing the client to report an error.
Therefore, you should explicitly return close when closing the connection.

Clients affected by this: okhttp3 <= 3.11.0

fatedier:
Do not checkout from master branch.

PR #1956, opened 2020-08-25
Run server plugins in consistent order

lukacu:
This fix should ensure that server plugins are always ordered in the same way when loading config, i.e. according to lexicographical order. This is important because in some cases plugins may expect that an action of some other plugin was already realized.


fatedier:
1. You'd better checkout from dev branch to avoid unnecessary commits.
2. It's ok and can we do more to let users specify the order of plugins? What's your opinion?

lukacu:
1. yes, I see. I mush have done something wrong, I do not open pull requests very often. I will correct it if it is possible.
2. I am not sure, this is the simplest way of specifying order by chaning the ids of the plugins. The other option is to change the way that config is loaded, then the order will be set explicitly in the config file.

fatedier:
This LGTM. You'd better create another PR base on dev branch.

PR #1974, opened 2020-09-08
Add tls configuration to both client and server

yuyulei:


models/config/client_common.go
```diff
@@ -320,5 +347,23 @@ func (cfg *ClientCommonConf) Check() (err error) {
 		err = fmt.Errorf("Parse conf error: invalid heartbeat_timeout, heartbeat_timeout is less than heartbeat_interval")
 		return
 	}
+
+	if cfg.TLSEnable == false {
+		if cfg.TLSCertFile != "" {
+			err = fmt.Errorf("Parse conf error: forbidden tls_cert_file, it only works when tls_enabled is true")
```
fatedier, line 353: Do not return error. Give it a warning like `tls_cert_file can only work when tls_enable is true`.

If users want to disable TLS, they can just comment `# tls_enable = true`.

fatedier (approved):

PR #2061, opened 2020-11-06
update ISSUE_TEMPLATE

BecodReyes:


fatedier:
I have a question, how github decide which langauge template to use?

BecodReyes:
> I have a question, how github decide which langauge template to use for different users?
> 
> And I'm prefer not to provide a Chinese version. It's not friendly for other peoples not familiar with Chinese. We'd better encourage our users communicate in English.

Thank you for your consideration for users who are not familiar with Chinese.

I haven't found a solution yet, so I created Chinese and English issue template. If you think this method is not good, I can delete the Chinese issue template.

fatedier:
Yes, just provide a English version.

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,5 @@
+blank_issues_enabled: false
+contact_links:
+  - name: DOCS
+    url: https://gofrp.org/en/docs/
```
fatedier, line 4: We haven't provide a English version here, just link repo README page?

BecodReyes: It was an oversight on my part, and I'm terribly sorry about it.

fatedier (approved):

PR #2073, opened 2020-11-12
Update issue template

BecodReyes:


BecodReyes:
Cancel and wait until finish writing the action before pull

PR #2098, opened 2020-11-23
Reconstruct config

yuyulei:


conf/frpc_full.ini
```diff
@@ -177,7 +177,7 @@ use_compression = true
 http_user = admin
 http_pwd = admin
 # if domain for frps is frps.com, then you can access [web01] proxy by URL http://test.frps.com
-subdomain = web01
+sub_domain = web01
```
fatedier, line 180: It's `subdomain` from parse code.

pkg/config/types.go
```diff
@@ -15,98 +15,508 @@
...
+	"gopkg.in/ini.v1"
 )
 
-type BandwidthQuantity struct {
-	s string // MB or KB
+// ClientCommonConf contains information for a client service. It is
+// recommended to use GetDefaultClientConf instead of creating this object
+// directly, so that all unspecified fields have reasonable default values.
+type ClientCommonConf struct {
+	auth.ClientConfig `ini:",,,,extends",json:"inline"`
```
fatedier, line 32: `go vet` failed

fatedier (changes requested):

pkg/config/client.go
```diff
@@ -0,0 +1,260 @@
...
+		return err
+	}
+
+	remotePorts, err := util.ParseRangeNumbers(remotePortStr)
+	if err != nil {
+		return err
+	}
+
+	if len(localPorts) != len(remotePorts) {
+		return fmt.Errorf("range section [%s] local ports number should be same with remote ports number", section.Name())
```
fatedier, line 230: Add section name as error message prefix where we called `appendTemplates`.

pkg/config/client.go
```diff
@@ -0,0 +1,260 @@
...
+			}
+			visitorConfs[prefix+name] = newConf
+		default:
+			return nil, nil, fmt.Errorf("section[%s] role should be 'server' or 'visitor'", name)
+		}
+	}
+	return proxyConfs, visitorConfs, nil
+}
+
+func appendTemplates(f *ini.File, section *ini.Section) error {
```
fatedier, line 210: We'd better use a more meaningful name like `renderRangeProxyTemplates`?

pkg/config/visitor.go
```diff
@@ -118,44 +93,24 @@ func (cfg *BaseVisitorConf) check() (err error) {
 	return
 }
 
-func (cfg *BaseVisitorConf) UnmarshalFromIni(prefix string, name string, section ini.Section) (err error) {
-	var (
-		tmpStr string
-		ok     bool
-	)
+func (cfg *BaseVisitorConf) decorate(prefix string, name string, section *ini.Section) error {
```
fatedier, line 96: It's releated to `ini`, use `UnmarshalFromIni` instead?

pkg/config/types.go
```diff
@@ -15,98 +15,508 @@
...
+}
+
+type SUDPProxySpec struct {
+	Role string `ini:"role",json:"role"`
+	Sk   string `ini:"sk",json:"sk"`
+}
+
+// Visitor
+var (
+	VisitorConfTypeMap = map[string]reflect.Type{
```
fatedier, line 486: Should this variable be exported?

pkg/config/types.go
```diff
@@ -15,98 +15,508 @@
...
+}
+
+// HTTP
+type HTTPProxyConf struct {
+	BaseProxyConf `ini:",,,,extends",json:"inline"`
+	HTTPProxySpec `ini:",,,,extends",json:"inline"`
+}
+
+type HTTPProxySpec struct {
+	DomainConf        `ini:",,,,extends",json:"inline"`
```
fatedier, line 402: Leave an empty line under embeded struct.

pkg/config/types.go
```diff
@@ -15,98 +15,508 @@
...
+	HTTPUser          string            `ini:"http_user",json:"http_user"`
+	HTTPPwd           string            `ini:"http_pwd",json:"http_pwd"`
+	HostHeaderRewrite string            `ini:"host_header_rewrite",json:"host_header_rewrite"`
+	Headers           map[string]string `ini:"-",json:"headers"`
+}
+
+// HTTPS
+type HTTPSProxyConf struct {
+	BaseProxyConf  `ini:",,,,extends",json:"inline"`
+	HTTPSProxySpec `ini:",,,,extends",json:"inline"`
```
fatedier, line 413: What's the reason of using too many nested embeded struct here?

```go
type HTTPSProxySpec struct {
    DomainConf `ini:",,,,extends",json:"inline"`
}
```

pkg/config/visitor.go
```diff
@@ -196,26 +165,40 @@ func (cfg *STCPVisitorConf) Compare(cmp VisitorConf) bool {
...
+	// Add custom logic validate, if exists
+
 	return
 }
 
-type XTCPVisitorConf struct {
-	BaseVisitorConf
-}
+// XTCP
+var _ VisitorConf = &XTCPVisitorConf{}
```
fatedier, line 201: Can we check if it satisfies `VisitorConf` interface in `init` functions using `VisitorConfTypeMap`?

So we don't need to write repeated content for each struct.

pkg/config/visitor.go
```diff
@@ -17,72 +17,47 @@ package config
...
-	Compare(cmp VisitorConf) bool
-	UnmarshalFromIni(prefix string, name string, section ini.Section) error
-	Check() error
-}
-
-func NewVisitorConfByType(cfgType string) VisitorConf {
-	v, ok := visitorConfTypeMap[cfgType]
+// DefaultVisitorConf creates a empty VisitorConf object by visitorType.
+// If visitorType doesn't exist, return nil.
+func DefaultVisitorConf(visitorType string) VisitorConf {
```
fatedier, line 26: TODO:

* NewVisitorConf()
* Unmarshal()
* Complete()
* Check()

pkg/config/proxy.go
```diff
@@ -528,112 +305,139 @@ func (cfg *HealthCheckConf) checkForCli() error {
...
+	cfg.BaseProxyConf.unmarshalFromMsg(pMsg)
+
+	// Add custom logic unmarshal if exists
+	cfg.RemotePort = pMsg.RemotePort
 }
 
-func (cfg *TCPProxyConf) UnmarshalFromIni(prefix string, name string, section ini.Section) (err error) {
-	if err = cfg.BaseProxyConf.UnmarshalFromIni(prefix, name, section); err != nil {
-		return
+func (cfg *TCPProxyConf) UnmarshalFromIni(prefix string, name string, section *ini.Section) error {
```
fatedier, line 335:
```go	
    err := section.MapTo(cfg)
    if err != nil {
        return err
    }

    err = cfg.BaseProxyConf.decorate(prefix, name, section)
    if err != nil {
        return err
    }
```

This code section is repeated in each ProxyConf. Can we wrap it as a function like

`preUnmarshalFromIni(cfg ProxyConf, prefix string,  name string, section *ini.Section)`?

pkg/config/visitor.go
yuyulei: actually, it is not a check in runtime,  but compile time，see: https://stackoverflow.com/questions/13194272/what-does-an-underscore-and-interface-name-after-keyword-var-mean

pkg/config/client.go
```diff
@@ -0,0 +1,260 @@
...
+		return ClientCommonConf{}, err
+	}
+
+	common := GetDefaultClientConf()
+	err = s.MapTo(&common)
+	if err != nil {
+		return ClientCommonConf{}, err
+	}
+
+	common.Metas = GetMapWithoutPrefix(s.KeysHash(), "meta_")
```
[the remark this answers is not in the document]
yuyulei: not recommend，It will require other developers to know `go-ini` lib.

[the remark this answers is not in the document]
fatedier: It's not elegant to split the unmarshal process into different steps.

We already use `go-ini` and it's flags. Put `ini` usage into one single README will help.

pkg/config/proxy.go
yuyulei: I has thinked about it, just like the demo `preUnmarshalFromIni(cfg ProxyConf,...)`, it is complicated when `cfg` is the interface but we want to edit `cfg.BaseProxyConf`, so I just keep it simple.

pkg/config/visitor.go
fatedier: I known. But it's worth nothing just increase the codes.

Each PR will run ci and e2e tests to make sure it's alright, this mistake won't be missed.

pkg/config/types.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2019 fatedier, fatedier@gmail.com
```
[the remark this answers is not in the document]
yuyulei: original copyright is in original `types.go`, which is renamed to `bandwidth.go`.

pkg/config/proxy.go
fatedier: `cfg.GetBaseInfo().decorate(...)`?

pkg/config/types.go
```diff
@@ -15,98 +15,508 @@
...
-	KB = 1024
+	"gopkg.in/ini.v1"
 )
 
-type BandwidthQuantity struct {
-	s string // MB or KB
+// ClientCommonConf contains information for a client service. It is
+// recommended to use GetDefaultClientConf instead of creating this object
+// directly, so that all unspecified fields have reasonable default values.
+type ClientCommonConf struct {
```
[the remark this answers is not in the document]
yuyulei: But if I change these struct defination with  `go-ini` or `json`, "all in one" is more convenient, so different situations diff.

sometimes, "all in one" is helpful to do somethings, like `protobuf` to auto code-generated.

fatedier (approved):

PR #2101, opened 2020-11-23
some dashboard refactor

hbrwang:
* upgrade vue-cli 4
* add eslint
* refactor requests

### Compiles and hot-reloads for development
```
npm run serve // not dev in vue-cli4
```

> 1. 目前vue-cli4是主流，vue-cli4环境无法运行vue-cli2项目。升级项目以便他人运行。
> 2. 使用 eslint 规范代码。


fatedier:
Can we just run `make build` to generate files?

We'd better auto install thrid party dependencies. Write the steps into `build.sh` if it's complicated.

fatedier:
A simple question: We haven't used vue-cli before, why we should import it? What's the benefits?

hbrwang:
> A simple question: We haven't used vue-cli before, why we should import it? What's the benefits?

vue-cli is more simpler and efficient. Using vue-cli is a more elegant way.

---

> Can we just run `make build` to generate files?
>
> We'd better auto install thrid party dependencies. Write the steps into `build.sh` if it's complicated.

I have't use `golang` and `make` before, but the `Makefile` is edited. I'm trying to do that now.

fatedier:
OK. Ping me later if you updated.

hbrwang:
Thank you for giving me the opportunity to learn `makefile`. It's amazing!

I ran `bash package.sh`,  and built files without error.

Then I tested the `windows_amd64` and `linux_amd64`. It seems all right.

_I didn't edit any files_

web/frps/Makefile
```diff
@@ -4,4 +4,4 @@ build:
 	@npm run build
```
fatedier, line 4: We should provide build scripts here to let others build it in a simple way.

I run `make build` on my machine and it failed. If it relies on other tools, we'd better descirbe it on README.

hbrwang: I just edited the `Makefile` to  `install` before `build`

fatedier: `yarn install` is used in README.md, what's the difference?

hbrwang: It was automatically generated and has now been edited

fatedier (approved):

fatedier:
@hbrwang Can you do the same refactor for frpc web UI?

hbrwang:
Sure, it's my pleasure

fatedier:
@hal-wang From this PR, the binary file size is much more greater than previous release.

I add `productionSourceMap: false` in `vue.config.js` to remove map files, but it still greater than before.

The web assets's size is 2MB now, before is 1.3MB.

I'm not sure if it's releated to vue-cli 4. If so, i will revert it to original state.

hal-wang:
It seems that vue-cli is larger than webpack, and a better solution is to use SDN. The PR #2129 ,assets's size is 0.6MB

fatedier:
@hal-wang It should not rely on any external resources. We compile all assets into binary files, so users can visit dashboard in any environment even without public network.

I will revert web assets to original state first. We can refactor later without making the web assets larger.

hal-wang:
@fatedier  OK, sorry for the trouble

fatedier:
@hal-wang I'm not familiar with front code. I'm confused why file size is grater than before that we add no new featrues.

Can we change to vue-cli without file size changed? If we ignored some configures?

hal-wang:
@fatedier There should be a problem with the configuration. Let me have a further study.

Issue #2102, opened 2020-11-23
Allow server plugin to talk to a https web service


mikehardenize:
**The solution you want**
The server plugin only seems to let you talk to an unencrypted http service, instead of a https one. I want to be able to talk to an encrypted web service, and to also be able to specify a custom CA or the ability to skip verification. E.g:

```
[plugin.user-manager]
addr = https://localhost:3000
path = /frp-auth
ops  = Login
tls_ca = /path/to/ca.pem
tls_verify = False
```

**Application scenarios of this function**
Our FRP server will live on a different machine to the one handling auth. We don't want to send unencrypted details over the wire.

PR #2103, opened 2020-11-23
Allow server plugin to talk to https services. Option for skipping tls verification

mikehardenize:
This is in relation to https://github.com/fatedier/frp/issues/2102. It allows me to use the following server plugin config to talk to a https service without tls verification:

```
[plugin.user-manager]
addr = https://localhost:8443
path = /login
ops  = Login
tls_verify = false
```

It is backwards compatible, so if you don't specify a schema, it will default to `http:` as before. If you don't specify `tls_verify` it will default to `true`.

I am not a golang programmer, so I apologise if the quality of my PR isn't good enough. I am happy to make changes if you need them.

### DIFF
--- a/pkg/config/server_common.go
+++ b/pkg/config/server_common.go
@@ -458,11 +458,16 @@ func UnmarshalPluginsFromIni(sections ini.File, cfg *ServerCommonConf) {
 	for name, section := range sections {
 		if strings.HasPrefix(name, "plugin.") {
 			name = strings.TrimSpace(strings.TrimPrefix(name, "plugin."))
+			var tls_verify, err = strconv.ParseBool(section["tls_verify"])
+			if err != nil {
+				tls_verify = true
+			}
 			options := plugin.HTTPPluginOptions{
-				Name: name,
-				Addr: section["addr"],
-				Path: section["path"],
-				Ops:  strings.Split(section["ops"], ","),
+				Name:      name,
+				Addr:      section["addr"],
+				Path:      section["path"],
+				Ops:       strings.Split(section["ops"], ","),
+				TlsVerify: tls_verify,
 			}
 			for i := range options.Ops {
 				options.Ops[i] = strings.TrimSpace(options.Ops[i])
--- a/pkg/config/server_common.go
+++ b/pkg/config/server_common.go
@@ -467,7 +467,7 @@ func UnmarshalPluginsFromIni(sections ini.File, cfg *ServerCommonConf) {
 				Addr:      section["addr"],
 				Path:      section["path"],
 				Ops:       strings.Split(section["ops"], ","),
-				TlsVerify: tls_verify,
+				TLSVerify: tls_verify,
 			}
 			for i := range options.Ops {
 				options.Ops[i] = strings.TrimSpace(options.Ops[i])
--- a/pkg/plugin/server/http.go
+++ b/pkg/plugin/server/http.go
@@ -17,19 +17,22 @@ package plugin
 import (
 	"bytes"
 	"context"
+	"crypto/tls"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"net/http"
 	"net/url"
 	"reflect"
+	"strings"
 )
 
 type HTTPPluginOptions struct {
-	Name string
-	Addr string
-	Path string
-	Ops  []string
+	Name      string
+	Addr      string
+	Path      string
+	Ops       []string
+	TlsVerify bool
 }
 
 type httpPlugin struct {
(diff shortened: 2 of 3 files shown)


pkg/config/server_common.go
```diff
@@ -458,11 +458,16 @@ func UnmarshalPluginsFromIni(sections ini.File, cfg *ServerCommonConf) {
...
 			options := plugin.HTTPPluginOptions{
-				Name: name,
-				Addr: section["addr"],
-				Path: section["path"],
-				Ops:  strings.Split(section["ops"], ","),
+				Name:      name,
+				Addr:      section["addr"],
+				Path:      section["path"],
+				Ops:       strings.Split(section["ops"], ","),
+				TlsVerify: tls_verify,
```
yuyulei, line 470: `TlsVerify` -> `TLSVerify`

pkg/plugin/server/http.go
```diff
@@ -40,10 +43,24 @@ type httpPlugin struct {
...
+	if options.TlsVerify == false {
+		tr := &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		}
+		client = &http.Client{Transport: tr}
+	} else {
+		client = &http.Client{}
+	}
+
+	var url = fmt.Sprintf("%s%s", options.Addr, options.Path)
```
yuyulei, line 56: how about adding `schema`, e.g.
```
if !options.TlsVerify {
  schema = "http"
} else {
  schema = "https"
}
...
url = fmt.Sprintf("%s://%s%s", schema, addr, path)
```

add I'll appreciate it very much if you could add [e2e test](https://github.com/fatedier/frp/blob/HEAD/test/e2e/plugin/client_plugins.go)

fatedier: `scheme` is not releated to `TLSVerify` since `TLSVerify` is only used for `HTTPS`.

mikehardenize: Yeah, what fatedier said. TLSVerify just allows us to skip certificate verification on a https connection. You'd probably not want that in a production setting, but it can be useful during development. It certainly is for me at least.

yuyulei: @mikehardenize @fatedier try 
```
[common]
bind_port = 7000

[plugin.user-manager]
addr = localhost:9000
path = /handler
ops = Login
tls_verify = false
```
some errer will happen, because the url(http://localhost:9000) not match tls config(skip: true). So I think it better to build the "suitable schema"(not "http" always) by tls_verify when schem not exists.

mikehardenize: Are you saying that the existence of a `tls_verify` config item should dictate whether or not we use https as the scheme and that we should never accept a scheme in the value for `addr`? I guess it *could* work that way, but why is that way better than the way I already did it?

fatedier: It's nothing about `TLSVerify`.  We can't detect if it's HTTPS server by this.

1. Default schema is `HTTP`.
2. If schema is `HTTPS`, build TLS transport with `TLSVerify`.

pkg/plugin/server/http.go
```diff
@@ -40,10 +43,24 @@ type httpPlugin struct {
 }
 
 func NewHTTPPluginOptions(options HTTPPluginOptions) Plugin {
+	var client *http.Client
+	if options.TLSVerify == false {
```
fatedier, line 47:
```go
var url = fmt.Sprintf("%s%s", options.Addr, options.Path)
if !strings.HasPrefix(url, "https://") && !strings.HasPrefix(url, "http://") {
    url = "http://" + url
}

var client *http.Client
if strings.HasPrefix(url, "https://") {
    tr := &http.Transport{
        TLSClientConfig: &tls.Config{InsecureSkipVerify: !options.TLSVerify},
    }
    client = &http.Client{Transport: tr}
} else {
    client = &http.Client{}
}
```

pkg/plugin/server/http.go
yuyulei: @mikehardenize maybe what I wrote confuses you. Now I reorganize it:

1. if `addr` is not specified the schema(`http` or `https`), we will use default schema `http`. Then follow 2.
2. if the schema of `addr` is `http`, no matter what `tls_verify ` value is, we will use default transport. That mean `tls_verify` will not decide which transport to use. 
3. if and only if  the schema of `addr` is `https` and `tls_verify` value is `false`, now, we will use transport with tlsconfig(InsecureSkipVerify: true).

what do you think ?

mikehardenize: Ok. I think I understand now. Although I'm not sure what practical difference it makes? I've just pushed up a commit with that change.

yuyulei: image

I  hope case#4 & case$6 can work. I think it "practical difference".

I would appreciate it if you can rebase commits.

mikehardenize: I don't know if you're asking me to change anything. Also, I don't know what you want me to rebase against as there are no new commits since mine. I think there might be a language barrier that is preventing this from progressing. Would it make sense for you to make any required changes, rather than trying to get me to make them at this point? Alternatively you could just click approve and then merge and we can both move on?

fatedier (approved):

PR #2113, opened 2020-11-26
add launchd frpc  config for mac users

quaeast:
add a launchd config template file to /config/launchd.


fatedier:
I prefer put this deploy scripts or files to this repo https://github.com/gofrp/deploy

However, i don't have enough time for that now, maybe it will delay for some time.

PR #2129, opened 2020-12-04
Merge the web code for frps and frpc

hal-wang:
1. Merge the web code for frps and frpc
2. use the vue-element-admin

---

**Compiles and hot-reloads for development**

```shell
# frps
npm run serve:s

# frpc
npm run serve:c
```

**Compiles and minifies for production**

```shell
# frps
npm run build:s

# frpc
npm run build:c
```

---

The compiled files are in `./web/dist/frps` and `./web/dist/frpc`

screenshot


fatedier:
I'm not sure why we should import `vue-element-admin`.

I'm not familiar with front develop. From my point of view, I prefer not rely on many other frameworks. Do we really need a complicated framework since it's just a simple dashboard ?

hbrwang:
To be exact, it is `vue-admin-template`, and only a part of the necessary source code is used here. There is not too much code added, but it will be more organized when you add web functions in the future.

web/vue.config.js
```diff
@@ -0,0 +1,59 @@
...
+
+function getProxyTargetPort() {
+  switch (appType) {
+    case 'frps':
+      return 8081
+    case 'frpc':
+      return 8082
+    default:
+      return 8080
+  }
```
yuyulei, line 16: Actually, this requires frpc & frps to listen the specified port as dashboard's port, so could you make a change:
`8081 -> 7500 & 8082 -> 7400`, because some users set config file from [frpc_full.ini]( https://github.com/fatedier/frp/blob/dev/conf/frpc_full.ini#L52) or [frps_full.ini](https://github.com/fatedier/frp/blob/dev/conf/frps_full.ini#L36).

hal-wang:
- frpc 785 KB
- frps 1.16 MB

yuqianma:
I'm interested in this pr.
The file size is always a concern to me as I puts fprs in an OpenWRT router.
And I'd never think that the dashboard file's size is such big (>1M). Since there are few functions.

However, gzipped size is fairly small, around 400 kB.
Can anyone tell me the frontend files' size in frp binary is its original size or compressed?

But I still think there are too many redundant files inside the bundle:
image


Additional questions

First, is file size a priority thing ?

Second, what frontend features will frp dashboard need in a near future?
For example, only pie and column chart ? The UI components used now is enough? How many browsers do we support?

If there will be more sophisticated frontend features, optimization would be meaningless. Or I would like to have a try and see how small can we get in a current codebase.

fatedier:
@yuqianma
1. It's not compressed with [statik](https://github.com/rakyll/statik)
2. We prefer to keep the balance between file size and featrues. Reduce web assets < 1MB is better.
3. I have no more free time to focus on dashboard since this is not a front-end project. Dashboard is aim to provide a convenient way to manage frp. If users want more features, there could be another single web project backed with frp's HTTP API.

yuqianma:
> @yuqianma
> 
> 1. It's not compressed with [statik](https://github.com/rakyll/statik)
> 2. We prefer to keep the balance between file size and featrues. Reduce web assets < 1MB is better.
> 3. I have no more free time to focus on dashboard since this is not a front-end project. Dashboard is aim to provide a convenient way to manage frp. If users want more features, there could be another single web project backed with frp's HTTP API.

Oh, I forgot HTTP API.
Great. Perhaps I'll have a try this weekend.

yuqianma:
@fatedier What browsers we intend to support?

fatedier:
@yuqianma 
Chrome, Firefox, Safari

hal-wang:
> @yuqianma
> 
> 1. It's not compressed with [statik](https://github.com/rakyll/statik)
> 2. We prefer to keep the balance between file size and featrues. Reduce web assets < 1MB is better.
> 3. I have no more free time to focus on dashboard since this is not a front-end project. Dashboard is aim to provide a convenient way to manage frp. If users want more features, there could be another single web project backed with frp's HTTP API.

If web open gzip, frps 422kb, frpc 280kb. But this needs server support. `nginx` can do it, but I don't know `go`.

yuqianma:
> @yuqianma
> Chrome, Firefox, Safari. Maybe 90% frp users using chrome.

Ok, just in case of IE... You know.

yuqianma:
After some attempts this afternoon, I admit that it exceeds the size I estimated.
Take frs for example, it's still around 1M by simple tweaks. Or we'd have to revise much more. Which I think is not better than a rewrite.

I attach screenshots of tries in the end.
The only effect simple tweak is the last one, to comment out `import 'element-ui/lib/theme-chalk/index.css'` in `element-ui.js`.
Other screenshots are trying to figure out the minimal size we can get currently.

Some ideas if someone have time to do more:
1. Replace `Echarts` with [@antv/f2](https://bundlephobia.com/result?p=@antv/f2@3.8.1)
Echarts and Zrender get `376kB`. While `f2`'s whole size is `167.9kB`.

2. Or we can use other tech stacks? Don't know how's size going if we use vue3 and new element-ui. I personally like [preact](https://bundlephobia.com/result?p=preact@10.5.12) (10Kb unzipped) + tailwindcss for simple project. [svelte](https://svelte.dev/) is also ideal for small size.

3. More practical idea. Web files served by frp now are gzipped. So can we package gzipped files directly in the binary? Nginx can serve .gz files. Maybe we can get the same result.

截屏2021-01-31 18 02 25
截屏2021-01-31 18 02 51
---
截屏2021-01-31 18 04 08
截屏2021-01-31 18 04 27
---
截屏2021-01-31 18 08 04
截屏2021-01-31 18 09 52
---
截屏2021-01-31 18 12 53
截屏2021-01-31 18 13 44
---
截屏2021-01-31 18 17 01
截屏2021-01-31 18 17 27

PR #2178, opened 2021-01-06
Add close issue bot

BecodReyes:
Automatically close the issue that does not conform to the specification

Time is precious. We can't waste time on non conforming issues

Tips: After processing the pull request, you need to install the [close-issue-app](https://github.com/apps/close-issue-app) to make it work properly

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,5 +1,5 @@
 blank_issues_enabled: false
 contact_links:
   - name: DOCS
-    url: https://github.com/fatedier/frp
+    url: https://github.com/fatedier/frp/blob/dev/README.md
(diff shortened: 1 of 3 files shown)


fatedier:
Can we use github actions instead of an app?

BecodReyes:
It can be achieved, but it may take a long time to wait

fatedier:
No problem and thanks for you contribution. I prefer use github actions so all similar configures can be modified in this repo files, no need to find how to config it in other place.

BecodReyes:
The configuration file for this github app is also stored in this repo file.

You configure it just as you configure github action.

I've configured and tested it, and it will work after merging.

fatedier:
The configures include if we enable this feature.

BecodReyes:
What? I mean I've configured and tested it, out of the box.

PR #2198, opened 2021-01-18
fix: websocket on http

gao270615179:
sending websocket request on http port with `http` type, frps will forward request and add the header that including `Connection: close`, then ws service can't response normally.

issue is #2195 

set DisableKeepAlives=false


fatedier:
Thanks for your contribution and find the problem.

But we can't simplely modify the `transport.DisableKeepAlives` here since it will affect other normal HTTP requests.

I will try to fix it later.

fatedier:
@gao270615179 I have submit a new PR https://github.com/fatedier/frp/pull/2199

Would you have time to review on it and test if it fixed the websocket issue?

gao270615179:
fix on #2199

PR #2201, opened 2021-01-18
initial commit for https2https plugin

kevtainer:
this creates a client side plugin which will allow for a https vhost to communicate with a https backend


fatedier:
What's the scenario to use this plugin?

kevtainer:
off-loading certificates and secure communication to the frontend vhost functionality (similar to https2http), but instead of pointing at a `http` backend, points to a `https` backend, which ensures end-to-end encryption, and proper functionality of redirects on the backend.

yuyulei (approved):

PR #2218, opened 2021-01-25
Updated README

asheroto:
- General spelling corrections
- Improved TLS section

### DIFF
--- a/README.md
+++ b/README.md
@@ -541,8 +541,6 @@ tls_trusted_ca_file = ca.crt
 
 You will need **a root CA cert** and **at least one SSL/TLS certificate**. It **can** be self-signed or regular (such as Let's Encrypt or another SSL/TLS certificate provider).
 
-If you want to generate the needed certs fast, download and run [Generate-Certs](https://github.com/asheroto/Generate-Certs) and it will generate the certs for you in less than a minute. Supports Windows + Linux.
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -515,11 +516,34 @@ use_compression = true
...
+tls_only = true
+tls_enable = true
+tls_cert_file = certificate.crt
+tls_key_file = certificate.key
+tls_trusted_ca_file = ca.crt
+```
+
+You will need **a root CA cert** and **at least one SSL/TLS certificate**. It **can** be self-signed or regular (such as Let's Encrypt or another SSL/TLS certificate provider).
+
+If you want to generate the needed certs fast, download and run [Generate-Certs](https://github.com/asheroto/Generate-Certs) and it will generate the certs for you in less than a minute. Supports Windows + Linux.
```
yuyulei, line 544: Worrying about the validity of external resources(links, repos, etc.), We tend to all documents, image, etc. can be concentrated in the main repos(frp, [frp-doc](https://github.com/gofrp/frp-doc)).

We have [docs](https://gofrp.org/docs/features/common/network/network-tls/#openssl-%E7%94%9F%E6%88%90%E8%AF%81%E4%B9%A6%E7%A4%BA%E4%BE%8B) about how to generate tls certificates, and we will make this section of "generating certificates" consistent with it later.

Really thanks, and can you remove this line?

PR #2253, opened 2021-02-22
Update Makefile.cross-compiles

chinaboard:
support apple M1

### DIFF
--- a/Makefile.cross-compiles
+++ b/Makefile.cross-compiles
@@ -7,6 +7,8 @@ all: build
 build: app
 
 app:
+	env CGO_ENABLED=0 GOOS=darwin GOARCH=arm64 go build -trimpath -ldflags "$(LDFLAGS)" -o ./release/frpc_darwin_arm64 ./cmd/frpc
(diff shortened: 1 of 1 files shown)


Makefile.cross-compiles
```diff
@@ -7,6 +7,8 @@ all: build
 build: app
 
 app:
+	env CGO_ENABLED=0 GOOS=darwin GOARCH=arm64 go build -trimpath -ldflags "$(LDFLAGS)" -o ./release/frpc_darwin_arm64 ./cmd/frpc
```
fatedier, line 10: Should add if-else case only for go version >= v1.16

chinaboard: fixed

fatedier: `sed` options is not compitable with macos.

Maybe you can reset to last version and i will improve it later.

chinaboard: ``go  version | awk '{print $3}' | sed -r 's|^go([0-9.]+).*$|\1|' | awk -F. '{print $2}'``
tesed on macos 11.2(intel) and 11.2.1(m1)

PR #2279, opened 2021-03-06
Support reverseproxy to dashboard with additional parts in path

YK-Samgo:
/api/ -> ../api/
    To support reverseproxy to dashboard with additional parts in path.

The dashboard may fail to open with url like this:
    https://www.abc.com/frps/static/#/
Because **vendor.js** call **/api/serverinfo** as absolute path. In the example, https://www.abc.com/api/serverinfo
Replace **/api/** in **vendor.js** with **../api/** so that it call the api with a workable path. In the example above, it actually calls https://www.abc.com/frps/api/serverinfo, which can be correctly reverse proxyed.


fatedier:
Codes in `assets` directory is auto generated. You should modify codes in `web/frps`.

YK-Samgo:
OK. I will close this PR and perform another.

PR #2255, opened 2021-02-23
Let's get rid of ugly statik

fakeboboliu:
## Intro
go1.16 introduced the embed package, it's the more graceful solution for embedding file into binary.
https://golang.org/pkg/embed/

## Pros
 - The code itself is more graceful
 - It makes the binary smaller slightly

image
(Even with upx packed)

## Oh that's sweet, what's the cost?
Use go1.16, well i'm changed ci configurations already

PR #2289, opened 2021-03-10
Support reverseproxy to dashboard with additional parts in path.

YK-Samgo:
/api/ -> ../api/   in vendor.js
    Support reverseproxy to dashboard with addtional parts in path.

The dashboard may fail to open with url like this:
https://www.abc.com/frps/static/#/
Because vendor.js call /api/serverinfo as absolute path. In the example, https://www.abc.com/api/serverinfo
Replace /api/ in vendor.js with ../api/ so that it call the api with a workable path. In the example above, it actually calls https://www.abc.com/frps/api/serverinfo, which can be correctly reverse proxyed.


YK-Samgo:
By the way, statik is really stupid. It's obvious that data is zipped and then stored as string! Thus a single byte b'\x03' costs 4 bytes on disk. So **statik.go** is almost 3 times of the size of the zipped data.
image
image

fakeboboliu:
> By the way, statik is far from elegant. It's obvious that **data** is zipped and then stored as string! Thus a single byte **b'\x03'** costs 4 bytes on disk. So **statik.go** is almost 3 times of the size of the zipped data.

I replaced statik with embed in https://github.com/fatedier/frp/pull/2255, but it seems we should bear statik for a long time.

fatedier (approved):

PR #2304, opened 2021-03-16
config: inline is NOT SUPPORTED in encoding/json

fakeboboliu:
and why use it as the field name?


fatedier:
@fakeboboliu I have force-pushed dev branch before. Can you rebase current dev branch first and then force-push it?

fakeboboliu:
@fatedier done

fatedier (approved):

PR #2447, opened 2021-06-11
frps: allow to use custom status code for not found page/vhost

danielkucera:


danielkucera:
Hmm there is still something wrong:
```
$ curl localhost:8082 -v
*   Trying 127.0.0.1:8082...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8082 (#0)
> GET / HTTP/1.1
> Host: localhost:8082
> User-Agent: curl/7.68.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Date: Fri, 11 Jun 2021 10:23:49 GMT
< Content-Length: 83
< Content-Type: text/plain; charset=utf-8
< 
HTTP/1.0 200 OK
Content-Type: text/html
Server: frp/0.37.0
Content-Length: 0

* Connection #0 to host localhost left intact
```

fatedier:
404 is the meaning of `Not Found` for HTTP protocol, i don't want to let user specifiy this code.

danielkucera:
Why?

michaljemala:
Certain use-cases could find the ability to provide a custom status code quite useful. For instance, if frp server is used to route webooks, we can set it up to respond with 200 even though no client is connected. This would be a very handy feature to be used in dev phase.

fatedier:
> Certain use-cases could find the ability to provide a custom status code quite useful. For instance, if frp server is used to route webooks, we can set it up to respond with 200 even though no client is connected. This would be a very handy feature to be used in dev phase.

I don't want to provide a direct way to allow users change 404 NOT Found code to 200 OK. It's really strange and maybe confuse users.

How about support this feature by [Server Plugin](https://github.com/fatedier/frp/blob/dev/doc/server_plugin.md). It allows you write your own logic to control frps without mofidy core code.

We can add a new webhook like `HTTPNotFound`. It provides request's host and path, you can decide which code and body to response. Maybe somebody needs to return different codes for different domains.

Now server plugin only supports HTTP/HTTPS protocol, but we will support other plugin system future such as `lua` or `webassembly`.

PR #2475, opened 2021-07-09
fix: client services pull in network-online.target

jian-lin:
Client services should pull in network-online.target instead of
network.target according to the
[doc](https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/).


jian-lin:
bump

PR #2483, opened 2021-07-15
Honor `login_fail_exit` when reconnecting

jordemort:
`login_fail_exit` is honored when initially connecting, but currently has no effect when reconnecting.

This adds a check after attempting to reconnect, so that a failure there is treated the same as a failure during the initial connection, if `login_fail_exit` is set to true.

### DIFF
--- a/client/service.go
+++ b/client/service.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net"
+	"os"
 	"runtime"
 	"strconv"
 	"sync"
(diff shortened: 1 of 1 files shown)


jordemort:
Hm, this isn't quite right, let me see

jordemort:
I think this version of it does what I want, but I think I broke a test along the way

fatedier:
https://github.com/fatedier/frp/blob/dev/conf/frpc_full.ini#L68

```ini
# decide if exit program when first login failed, otherwise continuous relogin to frps
# default is true
login_fail_exit = true
```

jordemort:
Ok makes sense - would you consider a PR that added an additional option then? Our use case is that we have a server plugin that validates a time-limited token at login - at reconnect this token will be expired and invalid, so we want it to exit when login on reconnect fails so our scripts can restart it with a new token.

fatedier:
Can you submit another issue to describe all your feature request and we can discuss on that issue?

jordemort:
@fatedier I have opened https://github.com/fatedier/frp/issues/2487 to discuss, thanks.

PR #2541, opened 2021-08-21
Add support for HTTPS via ACME as a client plugin

MOZGIII:
This PR adds a new client plugin `https2http_acme` and an end-to-end test for it.
The end-to-end test requires a proper DNS record configuration, so it is skipped by default.


MOZGIII:
Why did the test fail?

MOZGIII:
It uses some kind of HTTP challenge (probably http-01), I didn't really read through the code but that must be the default. I really hacked this together rather quickly, so didn't spend a lot of time digging into the acme part.

In my testing, it for sure used an HTTP challenge, as expected. No need to reload, just run the frpcs, and it will tunnel the traffic to HTTP port via an HTTPS subdomain or custom domain, and it will keep the certificated up to date.

MOZGIII:
@fatedier can you merge this?

PR #2676, opened 2021-11-24
remove nil error

ChristLZS:


ChristLZS:
When new a service, a nil error will always be returned.

PR #2720, opened 2021-12-28
refactor: refine pkg net utils

bingtianbaihua:


pkg/util/net/conn.go
```diff
@@ -194,50 +195,61 @@ func ConnectServer(protocol string, addr string) (c net.Conn, err error) {
...
 	switch protocol {
 	case "tcp":
 		return gnet.DialTcpByProxy(proxyURL, addr)
-	case "kcp":
-		// http proxy is not supported for kcp
-		return ConnectServer(protocol, addr)
-	case "websocket":
-		return ConnectWebsocketServer(addr)
 	default:
 		return nil, fmt.Errorf("unsupport protocol: %s", protocol)
```
fatedier, line 229: Error message maybe updated to `fmt.Errorf("unsupport protocol: %s when connecting by proxy", protocol)`

pkg/util/net/dial.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+}
+
+func DefaultDialOptions() dialOptions {
+	return dialOptions{
+		protocol: "tcp",
+	}
+}
+
+func WithProxy(proxyURL string) DialOption {
```
fatedier, line 45: `WithProxy` -> `WithProxyURL`，params for `WithProxy` seems better if it's a new Proxy struct.

fatedier (approved):

Issue #2730, opened 2022-01-03
[security] Alpine version old , leading to Security vulnerabilities


nitinSophos:
### Bug Description

We are using Old Alpine Image 3.12.0 , latest is 3.15.0 
Old alpin image has multiple vulnerabilities including Libssl 1.1.1k , and LibreSSL. (This is not an exhaustive list)

Fix: Update alpine base image in both frpc and frps docker image.

### frpc Version

0.38.0

### frps Version

0.38.0

### System Architecture

linux/amd64

### Configurations

N/A

### Logs

N/A

### Affected area

- [ ] Docs
- [ ] Installation
- [ ] Performance and Scalability
- [X] Security
- [ ] User Experience
- [ ] Test and Release
- [ ] Developer Infrastructure
- [ ] Client Plugin
- [ ] Server Plugin
- [ ] Extensions
- [ ] Others

nitinSophos:
Raised PR: https://github.com/fatedier/frp/pull/2731

PR #2731, opened 2022-01-03
ISSUE: 2730 Alpine version update for security fixes

nitinSophos:
FRP Docker Images update alpine version to fix security vulnerabilties
Issue: https://github.com/fatedier/frp/issues/2730

### DIFF
--- a/dockerfiles/Dockerfile-for-frpc
+++ b/dockerfiles/Dockerfile-for-frpc
@@ -1,11 +1,11 @@
-FROM alpine:3.12.0 AS temp
+FROM alpine:3 AS temp
 
 COPY bin/frpc /tmp
 
 RUN chmod -R 777 /tmp/frpc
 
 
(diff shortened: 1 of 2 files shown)


dockerfiles/Dockerfile-for-frpc
```diff
@@ -1,11 +1,11 @@
-FROM alpine:3.12.0 AS temp
+FROM alpine:3.15.0 AS temp
```
fatedier, line 1: Can we just use `alpine:3` here and no need to always update it.

nitinSophos: @fatedier  yes `alpine:3` also works updating it

nitinSophos: @fatedier  updated, please merge and build when possible 👍

fatedier (approved):

PR #2802, opened 2022-02-16
Add audience to token generator

addisonj:
Certain implementations of oauth2, such as auth0, require the audience
parameter to be send to the token generation endpoint.

This commit adds an additional parameter to, by default, send the
audience parameter.

This should be be broadly compatible and ignored by other
implementations. Otherwise, we can consider an additional configuration
option


bingtianbaihua:
@addisonj   Can you show me some relative documents. I cannot search it in Internet.

fatedier:
Similar PR: https://github.com/fatedier/frp/pull/1714 @bingtianbaihua

bingtianbaihua:
> Similar PR: #1714 @bingtianbaihua

got

addisonj:
Close in favor of https://github.com/fatedier/frp/pull/2814

Issue #2821, opened 2022-03-02
[Feature Request] Notify server plugins when a proxy is closed.


chengyuhui:
### Describe the feature request

Currently, server plugins only receive `NewProxy` events, it would be great if we can send `CloseProxy` messages to them as well (e.g. to use with https://github.com/arugal/frp-notify and notify users when a proxy goes offline for a certain length of time).

If the maintainers think that this is worth implementing, I can prepare a PR for this.

### Describe alternatives you've considered

Use external services to ping frp proxies and send notifications when they are down, but this sometimes does not work for non-HTTP(S) services.

### Affected area

- [ ] Docs
- [ ] Installation
- [ ] Performance and Scalability
- [ ] Security
- [ ] User Experience
- [ ] Test and Release
- [ ] Developer Infrastructure
- [ ] Client Plugin
- [X] Server Plugin
- [ ] Extensions
- [ ] Others

fatedier:
I think it's ok, but maybe a little different with `NewProxy`.

We can discuss more in your PR later.

chengyuhui:
Closing this since #2823 has been merged.

PR #2823, opened 2022-03-02
Notify server plugins when a proxy is closed

chengyuhui:
This is an implementation of #2821 ,but currently does not work.

It seems that similar types of plugin requests are dispatched here, but after adding a `fmt.Printf("%+v\n", m)`, no message is sent to this channel when I press `Ctrl-C` in `frpc`, am I missing something?
https://github.com/fatedier/frp/blob/19739ed31a26994e1e68346324f83b9544f7a4b0/server/control.go#L455-L457

image
image

I have chosen to ignore any response from the plugin since there is nothing to modify in this operation (we can't prevent a proxy from closing).

Help is needed also for end-to-end tests, since it is passing when really should not.

### DIFF
--- a/pkg/plugin/server/manager.go
+++ b/pkg/plugin/server/manager.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/fatedier/frp/pkg/util/util"
 	"github.com/fatedier/frp/pkg/util/xlog"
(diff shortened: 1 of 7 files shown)


fatedier:
It's a bit more complicated.

We don't promise that frpc can always send msg.CloseProxy to frps before exiting. Instead, you should inject logic here https://github.com/fatedier/frp/blob/dev/server/control.go#L375 when we really close and release proxy.

And we should not block on server plugin calls for `CloseProxy` message.  It's better to send notify requests asynchronously. We don't need to wait for plugin's response. It's just a notify message.

For some special cases, such as network error between frpc and frps, it has risk to send all `CloseProxy` messages to plugins. It will create too many connections if there are lots of frpcs and proxies.

chengyuhui:
This now works based on my local tests and e2e tests.

> For some special cases, such as network error between frpc and frps, it has risk to send all CloseProxy messages to plugins. It will create too many connections if there are lots of frpcs and proxies.

I think that it's fine, and the plugin itself should handle this situation and aggregate (or delay) the notifications so that it does not trigger rate limits or annoy the user, we shouldn't handle this on our own.

fatedier:
> I think that it's fine, and the plugin itself should handle this situation and aggregate (or delay) the notifications so that it does not trigger rate limits or annoy the user, we shouldn't handle this on our own.

Since we use HTTP/1.1，frps will create new connections for each new concurrent request. It will consume many fds and there are also local ports limit (maybe 30000 - 60000) to the same plugin port.

It's nothing about plugin. frps will get network error in that scenario.

chengyuhui:
What about one `CloseProxy` event per `frpc` in `stoper`? We can have a `ProxyNames []string` field for this. Aggregating the shutdown of multiple `frpc`s seems difficult.

Since it is still possible for a client to close a single proxy by sending `CloseProxy` messages, we need to handle this case as well, put a single name in that array.

fatedier:
> What about one `CloseProxy` event per `frpc` in `stoper`? We can have a `ProxyNames []string` field for this. Aggregating the shutdown of multiple `frpc`s seems difficult.
> 
> Since it is still possible for a client to close a single proxy by sending `CloseProxy` messages, we need to handle this case as well, put a single name in that array.

It's not the best practice.

You can simply add some notes in the server plugin docs to warn others not to use this in such a scenario.

I don't want to cost too much time on this since we will focus on frp v2 in the future.

chengyuhui:
Fine, I have rolled back the changes and updated documentation.

pkg/plugin/server/manager.go
```diff
@@ -127,6 +132,27 @@ func (m *Manager) NewProxy(content *NewProxyContent) (*NewProxyContent, error) {
 	return content, nil
 }
 
+func (m *Manager) CloseProxy(content *CloseProxyContent) {
```
fatedier, line 135: `error` should be returned here.

fatedier: You can return the aggregate errors.

chengyuhui: Errors are now returned, but not handled at call site, since that is at most some duplicated logs.

fatedier (approved):

Issue #2831, opened 2022-03-10
Cant connect with tcp over tls, but non-tls tcp works


presidenten:
### tl;dr ###

**What I did**
I expose server through ingress with tcp - works
I expose server through ingress with tcp over tls and same port - `login to server failed: session shutdown` 

**What I expected to happen**
I expected frp to connect like normal

---

### Long version of setup and what I tried to verify that all parts are working ###

I have the frp server in a kubernetes cluster behind an ingress.

If I just expose frps trough tcp routing with sni '*', then frpc connects just fine.
<img width="1092" alt="image" src="https://user-images.githubusercontent.com/3883897/157559379-316249f7-d48b-4bc4-909c-2a769283411b.png">

But we need several frp instances in our cluster and want to do routing based on sni name, which requires tls.

So I got a certificates for a couple of subdomains from lets encrypt and tested with two tcp-echo containers, that got hosted on their own certificate and sni. Connecting with 
```bash
openssl s_client -servername echo-1.mydomain.com -connect echo-1.mydomain.com:32000
```
and
```bash
openssl s_client -servername echo-2.mydomain.com -connect echo-2.mydomain.com:32000
```
that both go through the same LB into the same ingress works as expected. I can type messages and get them echoed.
![image](https://user-images.githubusercontent.com/3883897/157558554-ca1dc8b4-4c48-43cb-8844-2067b6870e79.png)

But when I try to expose frps over tls things stop working. 
The ingress terminates tls, and forwards the traffic to frps on binding port 80.

Even though I can still see the frp 404 screen if I use https in address, instead of http, like so - https://frps-1.mydomain.com:32000
<img width="294" alt="image" src="https://user-images.githubusercontent.com/3883897/157558370-3c8c81db-c828-4369-9dac-dc75902cf7cd.png">

frpc refuses to connect.
<img width="749" alt="image" src="https://user-images.githubusercontent.com/3883897/157560661-5d37971e-1b1c-469e-ae5b-414093547e87.png">

Openssl still reports perfectly fine tls certificate from lets encrypt...
```bash
openssl s_client -servername frps-1.mydomain.com -connect frps-1.mydomain.com:32000
```
<img width="968" alt="image" src="https://user-images.githubusercontent.com/3883897/157561996-6ce59097-27f1-46be-9996-98e08c82e670.png">

---

Any ideas what needs to be configured or fixed to connect over tls with certificate termination in ingress?

### frpc Version

0.39.1

### frps Version

0.39.1

[...]

PR #2832, opened 2022-03-10
bugfix: Issue #2831 - Cant connect to frps behind ingress with tls

presidenten:
Fix for issue #2831 

`ServerName` gets added to `TLSConfig` even if no ca has been added, so that SNI based routing works as intended.

image

### DIFF
--- a/pkg/transport/tls.go
+++ b/pkg/transport/tls.go
@@ -100,6 +100,8 @@ func NewClientTLSConfig(certPath, keyPath, caPath, serverName string) (*tls.Conf
 		base.Certificates = []tls.Certificate{*cert}
 	}
 
+	base.ServerName = serverName
+
 	if caPath != "" {
 		pool, err := newCertPool(caPath)
 		if err != nil {
@@ -107,7 +109,6 @@ func NewClientTLSConfig(certPath, keyPath, caPath, serverName string) (*tls.Conf
 		}
 
 		base.RootCAs = pool
-		base.ServerName = serverName
 		base.InsecureSkipVerify = false
 	} else {
 		base.InsecureSkipVerify = true


pkg/transport/tls.go
```diff
@@ -100,14 +100,15 @@ func NewClientTLSConfig(certPath, keyPath, caPath, serverName string) (*tls.Conf
 		base.Certificates = []tls.Certificate{*cert}
 	}
 
+	base.ServerName = serverName
+
 	if caPath != "" {
```
bingtianbaihua, line 105: So, are your certPath, keyPath and caPath all empty ?

bingtianbaihua:
Hello @presidenten, I will track this problem.  I want to discuss with you.

First I have no kubernetes cluster, so I build a TLS proxy in my local and try to reproduce this problem.
The same with your config. But it does work.

So I'm curious how you found out, is there any rationale to justify this problem?


golang TLS Client config comments
```
// ServerName is used to verify the hostname on the returned
// certificates unless InsecureSkipVerify is given. It is also included
// in the client's handshake to support virtual hosting unless it is
// an IP address.
ServerName string
```

fatedier (approved):

bingtianbaihua (approved):

presidenten:
@fatedier Thanks for merging!

---

@bingtianbaihua By using tls sni name it is possible to use an lb to route traffic to different frp servers behind same ip and port, even if they are strict tcp by checking sni name before tls handshake.

The rationale is if you have a large kubernetes cluster with alot of frp servers and you want to expose them behind a layer 4 capable ingress. 
When its setup properly you can have as many frp servers as you want in your cluster, and expose them on the same ip and port, as long as they correspond to a unique SNI name. So there is only need for one NLB loadbalancer in aws for example. 

Most loadbalancers also have support for this functionality themselves.

bingtianbaihua:
> @bingtianbaihua By using tls sni name it is possible to use an lb to route traffic to different frp servers behind same ip and port, even if they are strict tcp by checking sni name before tls handshake.
> 
> The rationale is if you have a large kubernetes cluster with alot of frp servers and you want to expose them behind a layer 4 capable ingress. When its setup properly you can have as many frp servers as you want in your cluster, and expose them on the same ip and port, as long as they correspond to a unique SNI name. So there is only need for one NLB loadbalancer in aws for example.
> 
> Most loadbalancers also have support for this functionality themselves, so its not limited to kubernetes. You can use it in normal server setups behind a loadbalancer as well.

OK, I get it. Thank you for your explaination.  @presidenten

Issue #2492, opened 2021-07-22
Configurable websocket PATH


osnosn:
Create a new config record for websocket PATH when working in WS mode.

New config record for server (frps),
> websocket_path = /myfrp
   or
  websocket_path = /myfrp/aa23
 ( require start with '/', no space, up to 16 characters )

New config record for client (frpc), when protocol = websocket
> websocket_path = /myfrp
( require start with '/', no space, up to 16 characters )

If this empty or not config, then use default path "/~!frp".

parsanoori:
Any progress on this?

PR #2836, opened 2022-03-11
Configurable websocket PATH

wwqgtxx:
implement https://github.com/fatedier/frp/issues/2492


fatedier:
We don't want to support it since websocket protocol is tricky in frp. It's not the core feature this project should be invloved in.

coonection (approved):

Issue #2842, opened 2022-03-14
sudp在连接不上的时候产生大量的日志


sola97:
### Bug Description

我为teamview配置了sudp的visitor，但是在电脑未启动teamview的时候每隔三秒一条日志刷屏

### frpc Version

0.40.0

### frps Version

0.40.0

### System Architecture

linux/amd64

### Configurations

[teamviewer_udp_visitor]
type = sudp
role = visitor
server_name = teamviewer_sudp
sk = 123456
bind_addr = 0.0.0.0
bind_port = 5938

### Logs

![image](https://user-images.githubusercontent.com/17388849/158214104-e9ec52c8-6d39-46e3-befc-174bbc2f6ce7.png)

![image](https://user-images.githubusercontent.com/17388849/158212827-607578e6-9165-4d17-aadc-4a229f3a4ea8.png)

### Steps to reproduce

1. 
2. 
3. 
...

### Affected area

- [ ] Docs
- [ ] Installation
- [ ] Performance and Scalability
- [ ] Security
- [ ] User Experience
- [ ] Test and Release
- [ ] Developer Infrastructure
- [ ] Client Plugin
- [ ] Server Plugin
- [ ] Extensions
- [ ] Others

PR #2844, opened 2022-03-15
feat: SUDP alway reconnect and print too much log when no data ready

bingtianbaihua:
https://github.com/fatedier/frp/issues/2842

### DIFF
--- a/client/visitor.go
+++ b/client/visitor.go
@@ -380,44 +380,30 @@ func (sv *SUDPVisitor) dispatcher() {
 	var (
 		visitorConn net.Conn
 		err         error
+
+		firstPacket *msg.UDPPacket
 	)
 
 	for {
-		// loop for get frpc to frps tcp conn
-		// setup worker
-		// wait worker to finished
-		// retry or exit
-
-		if len(sv.sendCh) > 0 {
-			visitorConn, err = sv.getNewVisitorConn() // connect to frps when data is ready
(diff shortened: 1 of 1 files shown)


fatedier (changes requested):
It's not the way to solve the problem.

client/visitor.go
```diff
@@ -393,12 +405,13 @@ func (sv *SUDPVisitor) dispatcher() {
 			default:
 			}
 
-			time.Sleep(3 * time.Second)
+			util.RandomSleep(3*time.Second, 0.9, 1.1)
```
fatedier, line 408: No need to sleep, just clear data in sendCh and wait for another retry.

client/visitor.go
```diff
@@ -377,12 +377,24 @@ func (sv *SUDPVisitor) Run() (err error) {
 func (sv *SUDPVisitor) dispatcher() {
 	xl := xlog.FromContextSafe(sv.ctx)
 
+	var (
+		visitorConn net.Conn
+		err         error
+	)
+
 	for {
 		// loop for get frpc to frps tcp conn
```
fatedier, line 386: `// Create a new work connection between frpc and frps when receiving new contents and there is no active work connection`

fatedier (changes requested):

fatedier (approved):

PR #2856, opened 2022-03-18
Golangci-lint

dzyanis:
There are many issues with formatting, misspellings and even potentially dangerous bugs.
Even standard `go vet` returns two warnings:
```bash
$ go vet ./...
# github.com/fatedier/frp/test/e2e/plugin
test/e2e/plugin/server.go:148:18: Sprintf call needs 1 arg but has 2 args
# github.com/fatedier/frp/cmd/frpc/sub
cmd/frpc/sub/root.go:125:2: misuse of unbuffered os.Signal channel as argument to signal.Notify
```

I'd recommend to integrate [golngci-lint](https://golangci-lint.run/) and then enable linters one by one.


fatedier:
I don't want to  use golangci-lint.

You can create pr for `go vet` issue or i will fix them later.

dzyanis:
> You can create pr for `go vet` issue or i will fix them later.

Sure, here another one [PR](https://github.com/fatedier/frp/pull/2858)

> I don't want to use golangci-lint.

Okay, up to you! But can you please explain why? It's very interesting for the [golangci-lint community](https://gophers.slack.com/archives/CS0TBRKPC). Maybe you have some good arguments that can help improve it.

PR #2858, opened 2022-03-19
Fixed go vet warnings

dzyanis:
Fixed go vet warnings:
```bash
$ go vet ./...
# github.com/fatedier/frp/test/e2e/plugin
test/e2e/plugin/server.go:148:18: Sprintf call needs 1 arg but has 2 args
# github.com/fatedier/frp/cmd/frpc/sub
cmd/frpc/sub/root.go:125:2: misuse of unbuffered os.Signal channel as argument to signal.Notify
```

### DIFF
--- a/test/e2e/plugin/server.go
+++ b/test/e2e/plugin/server.go
@@ -149,7 +149,7 @@ var _ = Describe("[Feature: Server-Plugins]", func() {
 			[tcp]
 			type = tcp
 			local_port = {{ .%s }}
-			remote_port = 0
+			remote_port = %d
 			`, framework.TCPEchoServerPort, remotePort)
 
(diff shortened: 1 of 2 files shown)


test/e2e/plugin/server.go
```diff
@@ -149,7 +149,7 @@ var _ = Describe("[Feature: Server-Plugins]", func() {
 			[tcp]
 			type = tcp
 			local_port = {{ .%s }}
-			remote_port = 0
+			remote_port = %d
 			`, framework.TCPEchoServerPort, remotePort)
```
fatedier, line 153:
```suggestion
			`, framework.TCPEchoServerPort)
```

`remote_port = 0` is  expected.

PR #2873, opened 2022-03-27
frpc: support multiple confs

fatedier:


cmd/frpc/sub/root.go
```diff
@@ -104,6 +103,31 @@ var rootCmd = &cobra.Command{
...
+				if d.IsDir() {
+					return nil
+				}
+				wg.Add(1)
+				go func() {
+					err := runClient(path)
+					if err != nil {
+						fmt.Printf("frpc service error for config file [%s]\n", path)
+					}
+					wg.Done()
```
bingtianbaihua, line 123: defer wg.Done() 

Maybe `runClient` has some panic and this code never be executed ?

fatedier: Program will exit if it panics, so there is no side effect. But using defer is ok.

bingtianbaihua:
LGTM

bingtianbaihua (approved):

PR #2913, opened 2022-04-26
Fix the systemd warning "Special user nobody configured, this is not …

ouliuquan:
…safe!" https://www.freedesktop.org/software/systemd/man/systemd.exec.html#DynamicUser=

### DIFF
--- a/conf/systemd/frpc.service
+++ b/conf/systemd/frpc.service
@@ -4,7 +4,7 @@ After=network.target
 
 [Service]
 Type=simple
-User=nobody
+DynamicUser=yes
 Restart=on-failure
 RestartSec=5s
 ExecStart=/usr/bin/frpc -c /etc/frp/frpc.ini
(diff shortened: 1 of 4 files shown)


blizard863:
I think there is no need.

I think "nobody" is a system default user, it has the least permissions and can only read and write files that all users can access. I personally find it safest to run frp as this user.

https://news.ycombinator.com/item?id=17721400

https://www.freedesktop.org/software/systemd/man/systemd.exec.html#User/Group%20Identity

https://askubuntu.com/questions/329714/what-is-the-purpose-of-the-nobody-user

https://wiki.ubuntu.com/nobody

https://news.ycombinator.com/item?id=17721400

ouliuquan:
@blizard863 Hi, thanks for searching so much information. I noticed maybe systemd isn't perfect. We should be more careful about everything. "nobody" seems to be better than "DynamicUser".

The PR will be closed in two days.

PR #2925, opened 2022-05-04
package for openbsd

0xfoura:
frp builds and runs just fine on OpenBSD so it could be packaged for it.

### DIFF
--- a/Makefile.cross-compiles
+++ b/Makefile.cross-compiles
@@ -2,7 +2,7 @@ export PATH := $(GOPATH)/bin:$(PATH)
 export GO111MODULE=on
 LDFLAGS := -s -w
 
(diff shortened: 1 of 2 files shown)


Makefile.cross-compiles
```diff
@@ -2,7 +2,7 @@ export PATH := $(GOPATH)/bin:$(PATH)
 export GO111MODULE=on
 LDFLAGS := -s -w
 
-os-archs=darwin:amd64 darwin:arm64 freebsd:386 freebsd:amd64 linux:386 linux:amd64 linux:arm linux:arm64 windows:386 windows:amd64 linux:mips64 linux:mips64le linux:mips:softfloat linux:mipsle:softfloat
```
fatedier, line 5: Not inclined to add `openbsd:386`.

You can remove `freebsd:386` by the way.

0xfoura: why not build for x86 bsd? i still have some 32-bit machines.

fatedier: You can build it by yourself. But we are tend to encourage upgrading to 64-bit architectures in release assets.

PR #2966, opened 2022-06-03
feat: support aead encryption

mzz2017:
As we all know, stream encryption is vulnerable and recognizable. Compared with that, AEAD is more difficult to break and is a mainstream encryption method.

Related PR: https://github.com/fatedier/golib/pull/16

We enables frp to support the AEAD method (currently only aes-gcm).

Notice: we should remove the `replace` instruction in `go.mod` before merging. It should be only for our test.

Example configs:

Server side:
```ini
[common]
bind_port = 7000
token = my_token
aead = true
log_level = info
```

Client side:
```ini
[common]
server_addr = 127.0.0.1
server_port = 7000
token = my_token
aead = true

[iperf3]
type = tcp
local_ip = 127.0.0.1
local_port = 5201
remote_port = 5202
use_encryption = true
use_aead = true
```
Then we can use following commands to test:
```bash
iperf3 -s -p 5201
go run ./cmd/frps -c /tmp/frps.ini
go run ./cmd/frps -c /tmp/frpc.ini
iperf3 -c localhost -p 5202
```

### DIFF
--- a/client/control.go
+++ b/client/control.go
@@ -284,7 +284,7 @@ func (ctl *Control) reader() {
 	defer ctl.readerShutdown.Done()
 	defer close(ctl.closedCh)
 
-	encReader := crypto.NewReader(ctl.conn, []byte(ctl.clientCfg.Token))
+	encReader := crypto.NewReader(ctl.conn, []byte(ctl.clientCfg.Token), ctl.clientCfg.Aead)
 	for {
 		m, err := msg.ReadMsg(encReader)
 		if err != nil {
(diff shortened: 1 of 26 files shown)


fatedier:
It's deprecated. Please use TLS instead.

mzz2017:
TLS is pretty good but also really inconvenient.

I believe many users do not like to configure TLS and so as my friends and me. I think it is a nice enhancement for token.

fatedier:
> TLS is pretty good but also really inconvenient.
> 
> I believe many users do not like to configure TLS and so as my friends and me. I think it is a nice enhancement for token.

Just set `tls_enable = true` in frpc.ini.

mzz2017:
Will frp auto-generate cert files?

fatedier:
> Will frp auto-generate cert files?

Yes

Issue #1103, opened 2019-02-27
frpc p2p模式退出后连接一直保持


Cyclonekid:
Issue is only used for submiting bug report and documents typo. If there are same issues or answers can be found in documents, we will close it directly.
(为了节约时间，提高处理问题的效率，不按照格式填写的 issue 将会直接关闭。)

Use the commands below to provide key information from your environment:
You do NOT have to include this information if this is a FEATURE REQUEST

**What version of frp are you using (./frpc -v or ./frps -v)?**

0.24.1

**What operating system and processor architecture are you using (`go env`)?**

Linux amd64  

OS  Redhat 6.9

**Configures you used:**

A
[common]
server_addr = ***.***.***.***
server_port = 16000
token = **********

[p2p_ssh_11]
type = xtcp
sk = 111111111
local_ip = 127.0.0.1
local_port = 22

B
[common]
server_addr = ***.***.***.***
server_port = 16000
token = **********

[ssh_visitor_11]
type = xtcp
role = visitor
server_name = p2p_ssh_11
sk = 111111111
bind_addr = 0.0.0.0
bind_port = 6000

**Steps to reproduce the issue:**
1.   A，B点对点连接使用SSh，ssh客户端端正常退出后，A机上的TCP和UDP一直保持ESTABLISHED，在A机上会遗留大量的点对点的连接，（每登陆一次就产生2条ESTABLISHED连接）B机断开后连接不能正常关闭，需要重启A的frpc才能清除
2.
3.

**Describe the results you received:**

**Describe the results you expected:**

**Additional information you deem important (e.g. issue happens only occasionally):**

复现率100%

**Can you point out what caused this issue (optional)**

fatedier:
本地无法复现。

Cyclonekid:
![image](https://user-images.githubusercontent.com/17697818/53775192-783ec000-3f2c-11e9-831c-be7224c2abe4.png)

我这产生的链接记录，客户端全部已经退出，但链接一直保持ES，除非重新启动A机的FRPC
打开debug日志后，日志只显示连接的日志，看不到有退出的相关日志
![image](https://user-images.githubusercontent.com/17697818/53775214-90164400-3f2c-11e9-9efa-9c64a0292222.png)

fatedier:
在某些类型设备后确实会有问题，增加一个心跳包的机制，差不多 5 - 10 秒左右会释放这些连接。

[...]

PR #2982, opened 2022-06-16
Server Dashboard SSL Support

EMREOYUN:
Added a simple if-else statement that checks TLSOnly mode enabled. If enabled, serves the dashboard in HTTPS mode. Otherwise, serves the dashboard in HTTP mode. 
If TLSCertFile and TLSKeyFile are not used in config but TLSOnly is enabled, the HTTPS dashboard does not loads.

Tested on Debian Linux and Windows Server 2012.

### DIFF
--- a/cmd/frps/root.go
+++ b/cmd/frps/root.go
@@ -62,6 +62,9 @@ var (
 	maxPoolCount      int64
 	maxPortsPerClient int64
 	tlsOnly           bool
+	dashboardTLSMode     bool
+	dashboardTLSCertFile string
+	dashboardTLSKeyFile  string
 )
 
 func init() {
--- a/cmd/frps/root.go
+++ b/cmd/frps/root.go
@@ -37,31 +37,31 @@ var (
 	cfgFile     string
 	showVersion bool
 
-	bindAddr          string
-	bindPort          int
-	bindUDPPort       int
-	kcpBindPort       int
-	proxyBindAddr     string
-	vhostHTTPPort     int
-	vhostHTTPSPort    int
-	vhostHTTPTimeout  int64
-	dashboardAddr     string
-	dashboardPort     int
-	dashboardUser     string
-	dashboardPwd      string
-	enablePrometheus  bool
-	assetsDir         string
-	logFile           string
-	logLevel          string
-	logMaxDays        int64
-	disableLogColor   bool
-	token             string
-	subDomainHost     string
-	tcpMux            bool
-	allowPorts        string
-	maxPoolCount      int64
-	maxPortsPerClient int64
-	tlsOnly           bool
+	bindAddr             string
+	bindPort             int
+	bindUDPPort          int
+	kcpBindPort          int
+	proxyBindAddr        string
+	vhostHTTPPort        int
+	vhostHTTPSPort       int
+	vhostHTTPTimeout     int64
+	dashboardAddr        string
+	dashboardPort        int
+	dashboardUser        string
+	dashboardPwd         string
+	enablePrometheus     bool
+	assetsDir            string
+	logFile              string
+	logLevel             string
+	logMaxDays           int64
+	disableLogColor      bool
+	token                string
+	subDomainHost        string
+	tcpMux               bool
+	allowPorts           string
+	maxPoolCount         int64
+	maxPortsPerClient    int64
+	tlsOnly              bool
 	dashboardTLSMode     bool
 	dashboardTLSCertFile string
 	dashboardTLSKeyFile  string
--- a/cmd/frps/root.go
+++ b/cmd/frps/root.go
@@ -91,6 +91,9 @@ func init() {
 	rootCmd.PersistentFlags().StringVarP(&allowPorts, "allow_ports", "", "", "allow ports")
 	rootCmd.PersistentFlags().Int64VarP(&maxPortsPerClient, "max_ports_per_client", "", 0, "max ports per client")
 	rootCmd.PersistentFlags().BoolVarP(&tlsOnly, "tls_only", "", false, "frps tls only")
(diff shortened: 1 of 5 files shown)


fatedier:
`TLSOnly`, `TLSKeyFile`, `TLSCertFile` isn't for dashboard, but just used in connections between frpc and frps.

EMREOYUN:
I'm aware of this but since it will be installed on a remote machine, most users will prefer SSL support on dashboard too. Should I implement new options on config for that?

fatedier:
OK. You can add configs with `dashboard` prefix.

EMREOYUN:
I made it into a different config variable, but the current problem is that the cert file cannot be opened by frps, even if I correctly located the file and it cannot check if it actually exists. I checked the entire code but couldn't find where `tls_cert_file` or `tls_key_file` processed and checked if it exists or not. Can you tell me where `tls_cert_file` and `tls_key_file` were processed and checked at?

@fatedier

fatedier:
I test it ok. I'm not sure what's your problem.

EMREOYUN:
> I test it ok. I'm not sure what's your problem.

Weird. In my test, the program does not read the file if it is different than `tls_cert_file` and `tls_key_file`. Which system did you use for testing?

fatedier:
You should cosntruct a tls listener and call `Server` instead of `ServerTLS`. So you can get the error in advance, not in a goroutine.

`tls.LoadX509KeyPair(certFile, keyFile)`

EMREOYUN:
`LoadX509KeyPair` has been added and HTTPS mode works well with working checks. You can test it now. 

Also, do we need to add an e2e check for HTTPS mode for `make alltest` or do we not?

EMREOYUN:
@fatedier Still waiting your response, in case you did not saw my latest commits.

server/dashboard.go
```diff
@@ -70,20 +71,45 @@ func (svr *Service) RunDashboardServer(address string) (err error) {
 		http.Redirect(w, r, "/static/", http.StatusMovedPermanently)
 	})
 
-	server := &http.Server{
-		Addr:         address,
-		Handler:      router,
-		ReadTimeout:  httpServerReadTimeout,
-		WriteTimeout: httpServerWriteTimeout,
+	if svr.cfg.DashboardTLSMode {
```
fatedier, line 74: We needn't fullfill all variables.

```go
    server := &http.Server{
        Addr:         address,
        Handler:      router,
        ReadTimeout:  httpServerReadTimeout,
        WriteTimeout: httpServerWriteTimeout,
    }
    ln, err := net.Listen("tcp", address)
    if err != nil {
        return err
    }

    if svr.cfg.DashboardTLSMode {
        cert, err := tls.LoadX509KeyPair(svr.cfg.DashboardTLSCertFile, svr.cfg.DashboardTLSKeyFile)
        if err != nil {
            return err
        }
        tlsCfg := &tls.Config{
            Certificates: []tls.Certificate{cert},
        }
        ln = tls.NewListener(ln, tlsCfg)
    }
    go server.Serve(ln)
    return
```

cmd/frps/root.go
```diff
@@ -62,6 +62,9 @@ var (
 	maxPoolCount      int64
 	maxPortsPerClient int64
 	tlsOnly           bool
+	dashboardTLSMode     bool
```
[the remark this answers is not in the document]
EMREOYUN: So what is the relation between `dashboardTLSMode     bool` and `make` and `go fmt` commands? I don't understand.

server/dashboard.go
EMREOYUN: Resolved. Also Tried to create `tls.NewListener` with the HTTP listener but since it tries to direct HTTPS connections to HTTP host which gives the following error: `Client sent an HTTP request to an HTTPS server.`

EMREOYUN: Now go-version-last gives an error but I couldn't figured out why...

EMREOYUN:
Read and Write timeouts did give error on CircleCI so I removed them from HTTPS mode

server/dashboard.go
fatedier: I'm not sure why you create a new http server for that. Just reuse the same one.

SSL termination has been down in tls listener.

EMREOYUN: ReadTimeout and WriteTimeout gives error on CircleCI, refer to my previous reply.

fatedier: I don't know what's your meaning.

You'd better solve the problem, not ignore the configuration.

cmd/frps/root.go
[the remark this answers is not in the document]
fatedier: No matter what you do, keep correct code formats in your way.

server/dashboard.go
EMREOYUN: Okay, in CircleCI run #1103, can you guide me what went wrong on it?

cmd/frps/root.go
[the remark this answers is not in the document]
EMREOYUN: What is wrong with dashboardTLSMode? Do you mean tab spacing or something else by keeping good code format?

server/dashboard.go
fatedier: I think it's a temporary error. Retry the test can fix it.

cmd/frps/root.go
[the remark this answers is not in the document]
fatedier: Yes. https://go.dev/blog/gofmt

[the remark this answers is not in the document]
EMREOYUN: A new commit has been added, directly copied from the output from `gofmt -w`. Can you check it?

server/dashboard.go
EMREOYUN: Simplified both HTTP and HTTPS into single server and CircleCI didn't throw error this time. You can check it.

fatedier (approved):

PR #3030, opened 2022-07-22
fix: can not close frpc

ChristLZS:
when I run a frpc service, and then close it immediately(or in other specific time),it can no exit gracefully.


fatedier:
In fact, there is no graceful shutdown method between frpc and frps. It's ok to just close frpc.

PR #3068, opened 2022-08-21
Added OpExit

lowk3v:
I added an option that supported the server plugin with the op is Exit. When the FRP client cannot connect to a server then the server will push a message to the FRP-notify plugin (version OpExit supported at https://github.com/lowk3v/frp-notify/releases/tag/OpExitSupported)
I would like to help someone

### DIFF
--- a/doc/server_plugin.md
+++ b/doc/server_plugin.md
@@ -94,6 +94,29 @@ Client login operation
 }
 ```
 
+#### Exit
+
+Client exit operation. When client cannot connect to server that will notify to admin.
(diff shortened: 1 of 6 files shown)


fatedier:
We don't expose the concept of client to users. A client has no consistent unique name.

I'm cautious to support it now.

fatedier:
> I use a metas field (like meta_hostname) to define the unique name of the client

It's not the common way for all users.

The problem is that we should not send all client infos in this message, but a unique name like `CloseProxy` message.

PR #3155, opened 2022-10-31
Set least privileged token permission for GitHub Actions

boahc077:
Other than `golangci-lint.yml`, all other GitHub Actions workflow files have elevated token permissions as demonstrated by the following workflow execution logs.
https://github.com/fatedier/frp/actions/runs/3334432128/jobs/5517384845#step:1:19

This PR mitigates this issue. In addition to this PR, you should consider setting the following permission so that new workflow files will default to read-only permissions
https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#setting-the-permissions-of-the-github_token-for-your-repository

Signed-off-by: Ashish Kurmi <akurmi@stepsecurity.io>

### DIFF
--- a/.github/workflows/goreleaser.yml
+++ b/.github/workflows/goreleaser.yml
@@ -3,6 +3,9 @@ name: goreleaser
 on:
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
   goreleaser:
     runs-on: ubuntu-latest
--- a/.github/workflows/build-and-push-image.yml
+++ b/.github/workflows/build-and-push-image.yml
@@ -9,6 +9,9 @@ on:
         description: 'Image tag'
         required: true
         default: 'test'
+permissions:
+  contents: read
+
 jobs:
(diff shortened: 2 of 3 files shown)


.github/workflows/goreleaser.yml
```diff
@@ -3,6 +3,9 @@ name: goreleaser
 on:
   workflow_dispatch:
 
+permissions:
```
fatedier, line 6: `goreleaser` will create release and upload assets. Does it need write permission?

boahc077: `goreleaser` seems to be using a separate GitHub token (being populated by the `GPR_TOKEN` GitHub secret):
https://github.com/fatedier/frp/blob/dev/.github/workflows/goreleaser.yml#L34

Hence the GitHub Actions generated token doesn't need any write permissions.

Issue #3177, opened 2022-11-17
[Feature Request] Allow for OIDC "scope" to be defined


mattfeury:
Hello,

Currently trying to use FRP with oidc and auth0. Things work well, but it appears that the oidc calls hardcode the "Scopes" to the "audience" parameter. https://github.com/fatedier/frp/blob/da51adc27670f495478f50634462954319e4fc14/pkg/auth/oidc.go#L105

In auth0, we have defined scopes, e.g. `users:read` etc that are substantially different from the audience. Attempting to use it with the sensible defaults fails because it's trying to validate a scope that doesn't exist. `Client has not been granted scopes {audience_value}`. It seems that conflating audience with scope is problematic, at least for auth0, so i'd rather be able to define those values separately. does anyone see issue with this? seems like it's really that one offending line.

### Describe alternatives you've considered

i've tried to use `oidc_additional_*` fields, but it appears that `oidc_additional_scope` can not be overwritten:

```
couldn't generate OIDC token for login: oauth2: cannot overwrite parameter "scope"
```

### Affected area

- [ ] Docs
- [ ] Installation
- [ ] Performance and Scalability
- [X] Security
- [ ] User Experience
- [ ] Test and Release
- [ ] Developer Infrastructure
- [ ] Client Plugin
- [ ] Server Plugin
- [ ] Extensions
- [ ] Others

fatedier:
cc @blizard863

blizard863:
When I test by auth0 service, the client OidcAudience field must be set empty. You can have a try.

I test by auth0 and authing service. The following configs can all pass.

# auth0
```
# frpc.ini
authentication_method = oidc
oidc_client_id = pEtG3mpBQ7******2i3KqoBOJ
oidc_client_secret = F3492WS5pG7w5F49I74Tq7c******LDvF3o6UOe
oidc_audience = "" // should be empty
oidc_token_endpoint_url = https://dev-***.us.auth0.com/oauth/token


# frps.ini
authentication_method = oidc
oidc_issuer = https://dev-***.us.auth0.com/
oidc_audience = https://dev-***.us.auth0.com/api/v2/

```

# authing
```
frpc

authentication_method = oidc
oidc_client_id = 6211e411e******a41819
oidc_client_secret = e71227a1dff04******5cb92fd8
oidc_audience = foobar // this config cannot be empty, but anything
oidc_token_endpoint_url = https://****.authing.cn/oidc/token


frps

authentication_method = oidc
[...]

PR #3192, opened 2022-11-29
Support OIDC scope parameter

mattfeury:
allow scope to be passed in for oidc credentials, maintaining backwards compatibility. this was discussed in #3177

### DIFF
--- a/pkg/auth/oidc.go
+++ b/pkg/auth/oidc.go
@@ -104,7 +104,7 @@ func NewOidcAuthSetter(baseCfg BaseConfig, cfg OidcClientConfig) *OidcAuthProvid
 	}
 
 	// Previous versions hardcoded the scope to audience,
-	// so for backwards compatability, use that if no scope is set
+	// so for backwards compatibility, use that if no scope is set
 	scope := cfg.OidcAudience
 	if cfg.OidcScope != "" {
 		scope = cfg.OidcScope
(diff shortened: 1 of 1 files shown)


blizard863:
# OK. I test it pass.

scope means the permission defined by ourself. Just like auth0 shows.
image


---

# And I also have some ideas.

1.  I think more better is scope field shoube be an array not just a string.

2. audience field is the identify the client id which is Identifier. But golang oidc package not support audience field clearly.  I think the better is 

"oidc_audience"  not append to scope filed. like this.
```
func NewOidcAuthSetter(baseCfg BaseConfig, cfg OidcClientConfig) *OidcAuthProvider {
	eps := make(map[string][]string)
	for k, v := range cfg.OidcAdditionalEndpointParams {
		eps[k] = []string{v}
	}

	eps["audience"] = []string{cfg.OidcAudience}

	tokenGenerator := &clientcredentials.Config{
		ClientID:       cfg.OidcClientID,
		ClientSecret:   cfg.OidcClientSecret,
		Scopes:         []string{cfg.OidcScope},
		TokenURL:       cfg.OidcTokenEndpointURL,
		EndpointParams: eps,
	}

	return &OidcAuthProvider{
		BaseConfig:     baseCfg,
		tokenGenerator: tokenGenerator,
	}
}
```

@mattfeury

blizard863:
Audience and Scope is different, I think set audience value to scope argument is wrong !

mattfeury:
@blizard863 thanks for the feedback. a few thoughts:

> 1. I think more better is scope field shoube be an array not just a string.

this sounds reasonable. how do you think we should model it in the yaml config? just comma separated?


regarding your other asks, i'm not sure what you're suggesting. it does feel that this PR should be limited to only changes to the scope. changes to audience and other parameters seem "out of scope" (har har) for #3177.

if i get #1 fixed, that seems like something worth merging, agreed?

blizard863:
You should fix the CI first.

mattfeury:
> You should fix the CI first.

@blizard863 done!

blizard863:
LGTM

blizard863 (approved):

PR #3209, opened 2022-12-14
Fix basic example server to join in readme

mhmxs:
As i understand `frps` creates the proxy on `remote_port` of `frpc`. In that case port `6000` would be open on server A. Architecture diagram tells the same.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,7 @@ Note that `local_port` (listened on client) and `remote_port` (exposed on server
 
   `./frpc -c ./frpc.ini`
 
-5. From another machine, SSH to server A like this (assuming that username is `test`):
+5. From another machine, SSH to server B via server A  like this (assuming that username is `test`):
 
   `ssh -oPort=6000 test@x.x.x.x`
 
--- a/README.md
+++ b/README.md
@@ -142,7 +142,7 @@ Note that `local_port` (listened on client) and `remote_port` (exposed on server
 
   `./frpc -c ./frpc.ini`
 
-5. From another machine, SSH to server B like this (assuming that username is `test`):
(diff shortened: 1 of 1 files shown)


fatedier:
> From another machine, SSH to server B like this (assuming that username is test)

It means which real server the user wants to connect(use ssh). Server A is a jumpserver, not the target server.

mhmxs:
@fatedier first of all thanks for the quick answer. Port `6000` is open on the jump server. I had to redesign my stuff because I suspected port `6000` would open on server `B` and tried to ssh on port `6000`, but in my experience, the port is available where `frps` is running which is server `A`. With that switch, your project is working perfectly. Thanks for all the hard work.

mhmxs:
Line: https://github.com/fatedier/frp/blob/dev/conf/frpc_full.ini#L156 also tells `remote port listen by frps`, and also the readme: `Start frps on server A:`, So i'm totally sure `ssh request` must go to server A. @fatedier 🙏

fatedier:
Yes. Maybe my wording is not very accurate.

It means you want to login to server B by ssh not A.

README.md
```diff
@@ -142,7 +142,7 @@ Note that `local_port` (listened on client) and `remote_port` (exposed on server
 
   `./frpc -c ./frpc.ini`
 
-5. From another machine, SSH to server B like this (assuming that username is `test`):
+5. From another machine, SSH to server A like this (assuming that username is `test`):
```
mhmxs, line 145:
```suggestion
5. From another machine, SSH to server B via server A  like this (assuming that username is `test`):
```

What do you hink @fatedier ?

fatedier: ok

fatedier (approved):

Issue #3085, opened 2022-09-01
[Feature Request]  can add  bandwidth_limit  in server side


freshgeek:
### Describe the feature request

[Feature Request]  can add  bandwidth_limit  in server  side
add   bandwidth_limit  in client side  is not useful，  client is not controlled  

### Affected area

- [ ] Docs
- [ ] Installation
- [ ] Performance and Scalability
- [X] Security
- [ ] User Experience
- [ ] Test and Release
- [ ] Developer Infrastructure
- [ ] Client Plugin
- [X] Server Plugin
- [ ] Extensions
- [ ] Others

Becods:
[#bandwidth-limit](https://github.com/fatedier/frp#bandwidth-limit)

fatedier:
An optional method it the [server plugin](https://github.com/fatedier/frp/blob/dev/doc/server_plugin.md) to override proxy `bandwidth_limit` in frps.

cbodonnell:
Opened https://github.com/fatedier/frp/pull/3271 to support setting `bandwidth_limit` from a server plugin.  This allows the server to control the limit for each proxy.

tecufanujacu:
> An optional method it the [server plugin](https://github.com/fatedier/frp/blob/dev/doc/server_plugin.md) to override proxy `bandwidth_limit` in frps.

Is it possible to receive a little help? I'm trying to do this from several days, I'm trying to understand how, I studied the server plugin documentation and I'm also trying to understand how available server plugins works. Unfortunalty I'm failing.

PR #3271, opened 2023-01-26
support bandwidth_limit set by server plugin

cbodonnell:
Allow setting the proxy `bandwidth_limit` from a server plugin.

Fixes: #3085


pkg/config/proxy.go
```diff
@@ -399,6 +400,7 @@ func (cfg *BaseProxyConf) unmarshalFromMsg(pMsg *msg.NewProxy) {
 	cfg.ProxyType = pMsg.ProxyType
 	cfg.UseEncryption = pMsg.UseEncryption
 	cfg.UseCompression = pMsg.UseCompression
+	cfg.BandwidthLimit, _ = NewBandwidthQuantity(pMsg.BandwidthLimit)
```
cbodonnell, line 403: will result in an empty bandwidth limit if `NewBandwidthQuantity(...)` returns an error

fatedier:
I misunderstood the previous question a little bit. It's more difficult to support on server side.

Your change can't take effect for all proxy types, udp & http? They don't call `HandleUserTCPConnection` function.

And it should not take effect both on client and server side. It's a bit of a waste of resources.

I need to consider more about this feature later.

cbodonnell:
> I misunderstood the previous question a little bit. It's more difficult to support on server side.
> 
> Your change can't take effect for all proxy types, udp & http? They don't call `HandleUserTCPConnection` function.
> 
> And it should not take effect both on client and server side. It's a bit of a waste of resources.
> 
> I need to consider more about this feature later.

Thank you for the reply @fatedier, I see what you mean for other proxy types, I can investigate that and update the PR.

As for client and server setting this and wasting resources, is that the same for encryption and compression?  If the client controls the server, then they know whether or not it will be set and can avoid wasting their own resources.  If they do not control the server, then there is no way of knowing for sure.

Thinking about the possibilities if the client _does not_ control the server, for example:
* If the server sets a 10MB limit, but a client sets a 5MB limit, then 5MB will be the limit - client is able to additionally control the lower limit, if necessary
* If the server sets a 10MB limit, but a client sets a 20MB limit, then 10MB will be the limit - server is able to enforce the upper limit

There would be some wasted resources in the above scenarios, but there is also some additional functionality and control that can be achieved by each separate party (client and server).

Let me know if I can help out in any way.

fatedier:
Maybe add a new parameter `bandwidth_limit_mode`, optional value are `client` and `server`, default is `client`.

server/proxy/proxy.go
```diff
@@ -287,6 +289,15 @@ func HandleUserTCPConnection(pxy Proxy, userConn net.Conn, serverCfg config.Serv
 	if cfg.UseCompression {
 		local = frpIo.WithCompression(local)
 	}
+
+	limitBytes := cfg.BandwidthLimit.Bytes()
+	if limitBytes > 0 {
+		limiter := rate.NewLimiter(rate.Limit(float64(limitBytes)), int(limitBytes))
```
fatedier, line 295: The limiter is in connection-level here. But in frpc is proxy-level.

cbodonnell: I'll see how I can keep it stored at the proxy level here too.

cbodonnell:
> Maybe add a new parameter `bandwidth_limit_mode`, optional value are `client` and `server`, default is `client`.

Is this a proxy-level config set by the client and can the server override in a plugin?

I will try some things out this week and update the PR.

fatedier:
> Is this a proxy-level config set by the client and can the server override in a plugin?

Yes.

pkg/config/proxy.go
cbodonnell: Same for `bandwidth_limit_mode` below.  Defaults to `"client"` even if there is an error, so we can ignore and take the default if we want.

pkg/config/types.go
```diff
@@ -120,3 +120,24 @@ func (q *BandwidthQuantity) MarshalJSON() ([]byte, error) {
...
+	BandwidthLimitModeServer BandwidthLimitMode = "server"
+)
+
+// Create a new BandwidthLimitMode from a string. Default to "client".
+func NewBandwidthLimitMode(s string) (BandwidthLimitMode, error) {
+	switch BandwidthLimitMode(s) {
+	case BandwidthLimitModeClient, BandwidthLimitModeServer:
+		return BandwidthLimitMode(s), nil
+	default:
+		return BandwidthLimitModeClient, errors.New("invalid bandwidth_limit_mode")
```
cbodonnell, line 137: returning an error here to help with validation when parsing from the config file.  will always default to `"client"` in these cases.

cbodonnell:
@fatedier I have updated the PR to address your comments and added `bandwidth_limit_mode` defaulting to `"client"`.

conf/frpc_full.ini
```diff
@@ -154,6 +154,8 @@ local_ip = 127.0.0.1
 local_port = 22
 # limit bandwidth for this proxy, unit is KB and MB
 bandwidth_limit = 1MB
+# where to limit bandwidth, can be 'client' or 'server', default is 'client'
+bandwidth_limit_mode = server
```
fatedier, line 158: Change to `client` here.

pkg/config/proxy.go
```diff
@@ -141,6 +141,10 @@ type BaseProxyConf struct {
 	// BandwidthLimit limit the bandwidth
 	// 0 means no limit
 	BandwidthLimit BandwidthQuantity `ini:"bandwidth_limit" json:"bandwidth_limit"`
+	// BandwidthLimitMode specifies whether to limit the bandwidth on the
+	// client or server side. Valid values include "client" and "server".
+	// By default, this value is "client".
+	BandwidthLimitMode BandwidthLimitMode `ini:"bandwidth_limit_mode" json:"bandwidth_limit_mode"`
```
fatedier, line 147: Just declear as `string`?

pkg/config/proxy.go
```diff
@@ -365,6 +370,15 @@ func (cfg *BaseProxyConf) decorate(prefix string, name string, section *ini.Sect
 		}
 	}
 
+	if bandwidthMode, err := section.GetKey("bandwidth_limit_mode"); err == nil {
+		cfg.BandwidthLimitMode, err = NewBandwidthLimitMode(bandwidthMode.String())
```
fatedier, line 374: Validate it in `checkForCli` and `CheckForSvr`.

String value doesn't need a custom decode logic.

pkg/config/proxy.go
cbodonnell: changed to string

conf/frpc_full.ini
cbodonnell: changed to `client`

pkg/config/proxy.go
cbodonnell: moved validation of `bandwidth_limit_mode` to `CheckForCli`

pkg/config/types.go
cbodonnell: this is now outdated

cbodonnell:
@fatedier updated PR for your comments in commit 7ec3149.  E2E tests have 3 failing, but I cannot see why.  Can you help look at that when you have time?

fatedier:
Please rebase the dev code since we have changed some code to support go1.20.

fatedier:
You can run `make e2e` in your local environment to reproduce and debug the errors.

cbodonnell:
> You can run `DEBUG=true make e2e` in your local environment to reproduce and debug the errors.

Thank you, @fatedier. I had missed adding the proxy config defaults for the sub commands.  I fixed that and added new flags for `bandwidth_limit` and `bandwidth_limit_mode` to set these on the cli.

Also ran `gci write ...` to get `golangci-lint` passing.

Locally, all of the e2e tests are now passing.

fatedier (approved):

tecufanujacu:
Hello @cbodonnell, is it possible to have a little example for limit bandwidth server side? I'm trying to write a plugin for this but without success.

cbodonnell:
> Hello @cbodonnell, is it possible to have a little example for limit bandwidth server side? I'm trying to write a plugin for this but without success.

@tecufanujacu I'll mention that you will need to set both the `bandwidth_limit` and `bandwidth_limit_mode` in the `content` field of the response to [`NewProxy`](https://github.com/fatedier/frp/blob/dev/doc/server_plugin.md#newproxy).

I don't have a full example, but these snippets show how you can do it:
```go
func handleNewProxy(w http.ResponseWriter, req *http.Request) {
  // ... parse metadata from the request

  // get the original `content` from the proxy metadata
  content, ok := metadata["content"].(map[string]interface{})
  if !ok {
    // ... handle no content
  }

  content["bandwidth_limit"] = "1250KB"
  content["bandwidth_limit_mode"] = "server"

  // ... any additional handler logic

  contentJSON, err := json.Marshal(content)
  if err != nil {
    // ... handle error
  }

  response := fmt.Sprintf("{\"reject\": false, \"content\": %s}", string(contentJSON))
  w.WriteHeader(http.StatusOK)
  w.Write([]byte(response))
}
```

tecufanujacu:
I tried until now using the function handleNewProxy (as [here](https://github.com/yhl452493373/frps-panel/blob/3f7dd930fa9b8b671b5c71ffac3dbb96e54fc5b2/pkg/server/controller/op.go#L57)), but without any success, it should be really a simple thing but I'm failing somewhere, unfortunatly I'm not good enough with golang.

I'd liked to write and share a simple server plugin similar to [this](https://github.com/Parmicciano/frp_plugin_allowed_ports) and of which I'm trying to reuse some of the code, a server plugin in which it is possible to write the bandwidth in a simple txt file, a similar plugin will make what you did usable by everyone.

Obviously at the moment I tried hardcoding the bandwidth directly in the plugin, but before continuing I have to better study golang and I have to correctly understand how frps internally works.

Anyway @cbodonnell, thanks a lot for your work and for your suggestions, I'll retry to work on the plugin that I have in mind as soon as possible, and if I will success then  I'll share it.

Obviously any furher help is higly appreciated.

PR #3275, opened 2023-01-28
add max_delay_time parameter to frpc

suiahae:
Add the `max_delay_time` parameter to frpc to avoid too frequent reconnection requests from frpc to frps and keep trying reconnection #3270

### DIFF
--- a/client/service.go
+++ b/client/service.go
@@ -168,7 +168,7 @@ func (svr *Service) Run() error {
 
 func (svr *Service) keepControllerWorking() {
 	xl := xlog.FromContextSafe(svr.ctx)
-	maxDelayTime := 20 * time.Second
+	maxDelayTime := time.Duration(svr.cfg.MaxDelayTime) * time.Second
 	delayTime := time.Second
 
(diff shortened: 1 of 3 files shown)


suiahae:
I don't know why, but that's how it is.

When I set the MaxDelayTime in https://github.com/suiahae/frp/blob/011422c087948732b4ad75fca09758b4e0a73085/pkg/config/client_test.go#L299 to 20, it can pass the test, but other values do not (eg MaxDelayTime: 19, 21 , 60). 

The failure information is as follows:

```
......
=== RUN   TestGetHTTPSHostname
--- PASS: TestGetHTTPSHostname (0.20s)
PASS
coverage: 4.7% of statements
ok      github.com/fatedier/frp/pkg/util/vhost  (cached)        coverage: 4.7% of statements
?       github.com/fatedier/frp/pkg/util/xlog   [no test files]
FAIL
make: *** [Makefile:35: gotest] Error 1
```

suiahae:
After this fix, max_delay_time can be set to a value other than the default (20) in ./pkg/config/client_test.go).

PR #3404, opened 2023-04-12
client: support websocket over TLS (wss://).

Dirbaio:
This adds support for using `protocol=websocket` and TLS at the same time in `frpc`.

The issue was the TLS handshake has to run before the websocket handshake. This required a small patch in golib: https://github.com/fatedier/golib/pull/20

Tested working with this config. The server is running `frps` behind an NGINX reverse proxy that terminates TLS.

```ini
[common]
server_addr = frp.example.com
server_port = 443
protocol = websocket
tls_enable = true
disable_custom_tls_first_byte = true
```

### DIFF
--- a/client/service.go
+++ b/client/service.go
@@ -455,7 +455,12 @@ func (cm *ConnectionManager) realConnect() (net.Conn, error) {
 	protocol := cm.cfg.Protocol
 	if protocol == "websocket" {
 		protocol = "tcp"
-		dialOptions = append(dialOptions, libdial.WithAfterHook(libdial.AfterHook{Hook: frpNet.DialHookWebsocket()}))
+		dialOptions = append(dialOptions, libdial.WithAfterHook(libdial.AfterHook{
+			// make sure to do the websocket handshake after the TLS handshake,
+			// to support websocket over TLS.
+			Priority: libdial.WithTLSConfigPriority + 1,
+			Hook:     frpNet.DialHookWebsocket(),
+		}))
 	}
 	if cm.cfg.ConnectServerLocalIP != "" {
 		dialOptions = append(dialOptions, libdial.WithLocalAddr(cm.cfg.ConnectServerLocalIP))


fatedier:
Could you please also provide the testing nginx configuration file?

Dirbaio:
It's a bit complex. I'm using kubernetes with the [`ingress-nginx`](https://github.com/kubernetes/ingress-nginx) ingress controller.

[Here](https://gist.github.com/Dirbaio/e97c5012d029778614e2556c94fd9b2c) are the kubernetes manifests for deploying. The ingress controller generates a huge and unreadable `nginx.conf` so I'm not going to post it here. However, any "TLS + websocket reverse proxy" `nginx.conf` should do, for example this one: https://stackoverflow.com/questions/12102110/nginx-to-reverse-proxy-websockets-and-enable-ssl-wss

client/service.go
```diff
@@ -455,7 +455,12 @@ func (cm *ConnectionManager) realConnect() (net.Conn, error) {
 	protocol := cm.cfg.Protocol
 	if protocol == "websocket" {
 		protocol = "tcp"
-		dialOptions = append(dialOptions, libdial.WithAfterHook(libdial.AfterHook{Hook: frpNet.DialHookWebsocket()}))
+		dialOptions = append(dialOptions, libdial.WithAfterHook(libdial.AfterHook{
+			// make sure to do the websocket handshake after the TLS handshake,
+			// to support websocket over TLS.
+			Priority: libdial.WithTLSConfigPriority + 1,
```
fatedier, line 461: This change has the potential to break the e2e case located at https://github.com/fatedier/frp/blob/dev/test/e2e/basic/client_server.go#L100. 

We would like to avoid making any incompatible changes.

In my opinion, `tls over websocket` is not equivalent to `wss`. 

Therefore, if we intend to support wss, it would be better to move it from the protocol to the proxy parameter.

Dirbaio: oh okay! I didn't realize "TLS over websocket" was actually a desired combination. 

Can you clarify which exact `frpc.ini` settings should result in "websocket over TLS"? I'll update the code to match that. Thanks!

ksingh7:
@Dirbaio  First of all thanks a lot for attempting to add support for wss in frp . Love frp but definitely need a solution for wss , after reading this PR , seems like this is exactly i was looking for.

If you don't mind , i would be happy to use your version of frp (that has support for wss) and deploy that to my (mini production) and send feedback to your work. Can you share me details on which branch to use / or you have frps/c executable (with wss) support backed-in.

Looking forward for any support on wss , tried of having insecure websocket connection and chrome crying loud with red warning of insecure certificates.

client/service.go
fatedier: Perhaps an alternative solution could be to add support for `wss` to the `protocol` parameter. 
If the user configuration is set to `websocket`, it will remain unchanged. However, if the configuration is set to `wss`, the websocket and tls order can be adjusted to implement wss.

WDYT?

fatedier:
Closed by https://github.com/fatedier/frp/pull/3503

PR #3528, opened 2023-07-14
use faster base64

JackDrogon:
### Summary

use faster base64 package "github.com/cristalhq/base64" to replace std

### DIFF
--- a/cmd/frpc/sub/reload.go
+++ b/cmd/frpc/sub/reload.go
@@ -15,13 +15,13 @@
 package sub
 
 import (
-	"encoding/base64"
 	"fmt"
 	"io"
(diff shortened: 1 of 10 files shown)


fatedier:
This is not a path that will be called frequently. I would recommend using the standard library to minimize maintenance costs.

JackDrogon:
> This is not a path that will be called frequently. I would recommend using the standard library to minimize maintenance costs.

Each packet call base64. It's a frequently call.

fatedier:
For udp proxy, I plan to adopt a new method to avoid base64 operations.

I hope to use the standard library as much as possible. From experience, this will avoid many issues.

PR #3723, opened 2023-10-24
Feature/retry access token

RobKenis:
### <samp>🤖 Generated by Copilot at d0e5852</samp>

This pull request adds retry support for OIDC authentication in frp clients. It defines a new `AuthRetryConfig` struct to store the retry parameters, and uses it in the `OidcAuthProvider`, `ClientCommonConfig`, and `AuthOIDCClientConfig` structs. It also modifies the `pkg/auth/oidc.go` and `pkg/config/v1/client.go` files to implement the retry logic and configuration parsing.

### WHY
When the OIDC provider is unavailable when a new token is requested, the client
is stuck for an indefinite state. This is only resolved by restarting the client or
the server.

With this feature, we want to achieve a scenario where a token is requested for a
set amount of retries when the OIDC provider is not available.

### DIFF
--- a/pkg/config/v1/common.go
+++ b/pkg/config/v1/common.go
@@ -16,6 +16,7 @@ package v1
 
 import (
 	"github.com/fatedier/frp/pkg/util/util"
+	"time"
 )
 
 type AuthScope string
@@ -114,4 +115,13 @@ type HTTPPluginOptions struct {
 
 type HeaderOperations struct {
 	Set map[string]string `json:"set,omitempty"`
+}
+
+type AuthRetryConfig struct {
+	// MaxRetries specifies the maximum number of retries to authenticate
+	// with frps. By default, this value is 0.
+	MaxRetries int `json:"maxRetries,omitempty"`
+	// RetryDelay specifies the delay between retries to authenticate with
+	// frps, in seconds. By default, this value is 1.
+	RetryDelay time.Duration `json:"retryDelay,omitempty"`
 }
--- a/pkg/auth/oidc.go
+++ b/pkg/auth/oidc.go
@@ -17,6 +17,7 @@ package auth
 import (
 	"context"
 	"fmt"
+	"github.com/fatedier/frp/pkg/util/xlog"
 	"time"
 
 	"github.com/coreos/go-oidc/v3/oidc"
@@ -31,6 +32,8 @@ type OidcAuthProvider struct {
 	additionalAuthScopes []v1.AuthScope
 
 	tokenGenerator *clientcredentials.Config
+
+	retryConfig v1.AuthRetryConfig
 }
 
 func NewOidcAuthSetter(additionalAuthScopes []v1.AuthScope, cfg v1.AuthOIDCClientConfig) *OidcAuthProvider {
@@ -54,16 +57,19 @@ func NewOidcAuthSetter(additionalAuthScopes []v1.AuthScope, cfg v1.AuthOIDCClien
 	return &OidcAuthProvider{
 		additionalAuthScopes: additionalAuthScopes,
 		tokenGenerator:       tokenGenerator,
+		retryConfig:          cfg.AuthRetryConfig,
 	}
 }
 
-func withRetries(retries int, fn func() (accessToken string, error error)) (accessToken string, err error) {
-	exponentialBackOff := time.Second * 1
+func withRetries(retries int, delay time.Duration, fn func() (accessToken string, error error)) (accessToken string, err error) {
+	xl := xlog.FromContextSafe(context.Background())
+	exponentialBackOff := time.Second * delay
 	for i := 0; i < retries; i++ {
 		accessToken, err = fn()
 		if err == nil {
 			return accessToken, nil
 		}
+		xl.Warn("Failed to generate OIDC token for login: %v. Retrying %i more times", err, retries-i)
 		time.Sleep(exponentialBackOff)
 		exponentialBackOff *= 2
 	}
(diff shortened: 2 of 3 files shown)


RobKenis:
@fatedier Apologies for making a PR without creating an issue first. I read the contributing section too late. I would like to get feedback on the idea of adding retries to retrieving an access token.

vvanouytsel:
Nice! This feature would also help me out.

fatedier:
> When the OIDC provider is unavailable when a new token is requested, the client
is stuck for an indefinite state. This is only resolved by restarting the client or
the server.

If login fails, the client itself will retry. `generateAccessToken` will be called again. I don't understand the significance of adding new retry logic.

BTW it is more appropriate to discuss code details in PR. You can create a new issue and link it to this PR, then discuss it in the issue.

pkg/config/v1/common.go
```diff
@@ -115,3 +116,12 @@ type HTTPPluginOptions struct {
 type HeaderOperations struct {
 	Set map[string]string `json:"set,omitempty"`
 }
+
+type AuthRetryConfig struct {
+	// MaxRetries specifies the maximum number of retries to authenticate
+	// with frps. By default, this value is 0.
```
blizard863, line 122: However, this is an incompatible modification, and if someone hasn't set this value, it will use the default value of 0, which will result in a failure to retrieve the OIDC token.

blizard863: I think you should not add a break-change code.

blizard863:
I believe this should be an optional feature as not everyone needs to retry. Therefore, we should maintain the original logic of making the request only once.

RobKenis:
> I believe this should be an optional feature as not everyone needs to retry. Therefore, we should maintain the original logic of making the request only once.

I agree. We can make the `AuthRetryConfig` optional and use the existing logic by default, so make the request once without retries.

RobKenis:
> > When the OIDC provider is unavailable when a new token is requested, the client
> > is stuck for an indefinite state. This is only resolved by restarting the client or
> > the server.
> 
> If login fails, the client itself will retry. `generateAccessToken` will be called again. I don't understand the significance of adding new retry logic.
> 
> BTW it is more appropriate to discuss code details in PR. You can create a new issue and link it to this PR, then discuss it in the issue.
> 
> cc @blizard863

When the initial login fails, the client will exist and the process restarts. The issue only exists when the first login succeeds but the following ones fail. Then the client does not exit, but also does not retry fetching a token

fatedier:
@RobKenis Have you conducted any testing? If so, this result is unexpected, and we need to fix the bug instead of introducing a new logic.

RobKenis:
> @RobKenis Have you conducted any testing? If so, this result is unexpected, and we need to fix the bug instead of introducing a new logic.

We are testing 2 scenarios:
- FRPS is restarted. The client reconnects so this is fine
- Our OIDC provider, Keycloak, restarts. Then the clients is stuck on the following logs until we restart the client of the server process

```
2023/10/26 09:38:01 [E] [control.go:155] [3ee37dcc6cd4c83f] StartWorkConn contains error: received different OIDC subject in login and ping. original subject: a00465c5-59fb-4759-b225-680fef23a325, new subject: b056f095-3954-43dc-b3a6-dbbe4c177b49
2023/10/26 09:38:14 [I] [control.go:252] [3ee37dcc6cd4c83f] control writer is closing
2023/10/26 09:38:14 [I] [visitor_manager.go:77] [3ee37dcc6cd4c83f] gracefully shutdown visitor manager
2023/10/26 09:38:14 [I] [service.go:214] [3ee37dcc6cd4c83f] try to reconnect to server...
2023/10/26 09:38:14 [I] [service.go:301] [3ee37dcc6cd4c83f] login to server success, get run id [3ee37dcc6cd4c83f]
2023/10/26 09:38:14 [I] [proxy_manager.go:150] [3ee37dcc6cd4c83f] proxy added: [test-data-connector]
2023/10/26 09:38:14 [I] [health.go:108] [3ee37dcc6cd4c83f] [test-data-connector] health check status change to success
2023/10/26 09:38:14 [I] [proxy_wrapper.go:234] [3ee37dcc6cd4c83f] [test-data-connector] health check success
2023/10/26 09:38:15 [I] [control.go:172] [3ee37dcc6cd4c83f] [test-data-connector] start proxy success
2023/10/26 09:38:44 [W] [control.go:299] [3ee37dcc6cd4c83f] error during ping authentication: couldn't generate OIDC token for login: oauth2: cannot fetch token: 503 Service Unavailable
Response: <html>
<head><title>503 Service Temporarily Unavailable</title></head>
<body>
<center><h1>503 Service Temporarily Unavailable</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

fatedier:
It's in the ping message, not login.

As I mentioned before, this is a bug that needs to be fixed, without introducing any new retry logic.

fatedier:
@blizard863 Do you have time to help fix and test this issue? In the `ctl.authSetter.SetPing` in the client, if it fails, it should continue to the next iteration instead of exiting directly. This will cause the entire message processing loop to exit.

blizard863:
> @blizard863 Do you have time to help fix and test this issue? In the `ctl.authSetter.SetPing` in the client, if it fails, it should continue to the next iteration instead of exiting directly. This will cause the entire message processing loop to exit.
> 
> In addition, OIDC and token are different here. The time required here is unpredictable, so it is necessary to start a new goroutine to handle it instead of synchronous processing.

OK，I will test this scenario.

Issue #1176, opened 2019-04-04
proposal - reload config file on save


YanB25:
## abstract
I'd like to contribute to this project by adding a new feature. I wish to support reloading the config files automatically on save.
## details
Users can enable this feature in `frpc.ini`
```
reload_on_save = true
```
and whenever the config file is written, a `reload` is automatically triggered.
### implementation details
I would use [https://github.com/fsnotify/fsnotify](fsnotify)(Cross-platform file system notifications for Go) to watch the config files and use `go's channel` to pass the information asynchronously.
## motivation
It's more natural for users (including me) to `ssh` to the client and change the config file. AdminUI is great but `ssh` is better.
A reload on save would be helpful.

fatedier:
I'm glad for your upcoming contribution.

Can we use `reload_on_update` instead of `reload_on_save` ?

The repo `https://github.com/fsnotify/fsnotify` maybe has some problem on macos like missing some events.

YanB25:
> Can we use `reload_on_update` instead of `reload_on_save` ?

Okay.

> The repo `https://github.com/fsnotify/fsnotify` maybe has some problem on macos like missing some events.

The repo has some problems causing *extra* events on files ([this issue](https://github.com/howeyc/fsnotify/issues/62)). Users may find extra reload triggered in the log file. I couldn't find a stable library to watch a file on MacOS currently.

If extra reloads on MacOS is not acceptable, maybe we have to close this issue.

fatedier:
Double events is ok. The problem is missing events.

You can try to test it or at least reload conf at a fixed interval ?

YanB25:
> You can try to test it or at least reload conf at a fixed interval ?

Yes, I can get the `last modified time` at a fixed interval and reload on necessary, if you prefer to use polling.

> The problem is missing events.

I go through the issues and only find [this one](https://github.com/fsnotify/fsnotify/issues/220) related to losing events. There's no further information about it. Only double events bug is confirmed.

BTW, I have MacOS environment and I can have a test about it if necessary.

fatedier:
@YanB25 No progress?

zyfdegh:
Any update?

fatedier:
This has been open for a long time — thanks for the suggestion. Since then, frpc has added a web dashboard with config editing, a reload API, and a store API for dynamic proxy/visitor management at runtime. These cover the original use case, so closing this.

PR #3728, opened 2023-10-27
feature: reload config file on save

NanamiNakano:
### <samp>🤖 Generated by Copilot at 0bd61cb</samp>

This pull request adds a feature to enable the client to automatically reload its configuration file when it changes. It introduces a new option `reloadOnUpdate` in the client configuration file and a new dependency `fsnotify` for file system notifications. It modifies the `Service` struct, the `ClientCommonConfig` struct, and the example configuration file.

### WHY
Implement issue https://github.com/fatedier/frp/issues/1176

### DIFF
--- a/client/service.go
+++ b/client/service.go
@@ -18,6 +18,9 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
+	"github.com/fatedier/frp/pkg/config"
+	"github.com/fatedier/frp/pkg/config/v1/validation"
+	"github.com/fsnotify/fsnotify"
 	"io"
 	"net"
 	"runtime"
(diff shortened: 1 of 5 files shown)


fatedier:
I don't want to introduce the dependency of fsnotify, this library may have compatibility issues in different os.

NanamiNakano:
I have tested on macOS, Windows and Linux and it looks like it works well on all of these operating systems. I didn't test it on Docker， if this dependency does cause issues, I would try to implement this without this library.

fatedier:
Currently, there is not much inclination to introduce additional complexity for this feature. This involves os system calls, not issues with the library itself.

The ability to configure will be considered for refactoring in the future. The framework itself will strive to be lightweight, ensuring compatibility, while complex capabilities will be implemented by external tools/plugins.

NanamiNakano:
So does this mean that this feature should be implemented by a plugin when an external plugin interface is provided in a later version? I didn't find any documentation or interface for developing external plugins for the current version of frp.

fatedier:
I have some simple ideas about refactoring in configuration management, but they are not perfect. There will likely be significant changes in this area in the future. Without clear specifications, I am currently not inclined to add any extra functionality.

Currently, after you have modified the configuration file, you can use the `frpc reload` command to reload the configuration file, or use the functionality provided on the web page to modify the configuration and reload. Compared to introducing the inotify API, I believe these are already sufficient.

You can also use another tool to watch for file updates and then call the reload interface of frpc. This combination may also be more in line with the philosophy of Unix systems.

NanamiNakano:
I got that，so this functionality should not be implemented in the frp ontology, but there is no good way to implement it as an external plugin either. Hopefully future v2 releases will provide support for third-party client-side plugins.

This PR should be closed

fatedier:
@NanamiNakano It's something like this. I had considered using "reload" every once in a while to avoid using fsnotify, but upon careful consideration, it doesn't seem elegant enough. It feels like over-engineering.

Another idea is to add an `edit` command that will call the command line editor, such as `vim`. After the editor exits, it will automatically trigger a reload.

However, I still doubt if this is really necessary? Perhaps there are other features and directions that are more worth investing time in.

PR #3737, opened 2023-10-31
Improve docker images

reneleonhardt:
### <samp>🤖 Generated by Copilot at 24e5b67</samp>

This pull request refactors and improves the docker-related files and workflows of the project. It adds a `Dockerfile` that builds a single image for both `frpc` and `frps`, using build arguments and multi-stage build. It also adds a `.dockerignore` file and a `.github/dependabot.yml` file, and simplifies the `.github/workflows/build-and-push-image.yml` file. It deletes the `dockerfiles/Dockerfile-for-frpc` and `dockerfiles/Dockerfile-for-frps` files, as they are no longer needed.

### WHY

Docker improvements:
* Improve security by running rootless
* Add standard labels
* Unify docker build into single file


fatedier:
I am not inclined to use dependabot as the risks outweigh the benefits.

I apologize, but I do not have the energy to review so many changes that are unrelated to the code. It might be more appropriate if you could first describe your expected changes and reasons in an issue, and then break them down into smaller, individual changes.

reneleonhardt:
Didn't want to cause stress, I will break them into 2 PRs without dependabot (which doesn't automerge, only suggests PRs automatically).

Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM golang:1.21 AS builder
+
+WORKDIR /building
+COPY . .
+
+ARG APP
+RUN make ${APP}
+
+FROM alpine:3.18 AS runtime
```
fatedier, line 9: Before, it was `FROM alpine:3`?

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+ARG APP
+RUN addgroup -g 1000 -S ${APP} && adduser -u 1000 -S ${APP} -G ${APP} --home /app \
+ && echo -e "#!/bin/sh\nexec /usr/local/bin/${APP} \$@" > /app/entrypoint.sh \
+ && chmod +x /app/entrypoint.sh
+
+FROM alpine:3.18
+
+ARG APP
+ARG TITLE
+LABEL org.opencontainers.image.authors="fatedier <fatedier@gmail.com>"
```
fatedier, line 20: Why do we need these labels? What benefits will it bring?

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+COPY . .
+
+ARG APP
+RUN make ${APP}
+
+FROM alpine:3.18 AS runtime
+
+ARG APP
+RUN addgroup -g 1000 -S ${APP} && adduser -u 1000 -S ${APP} -G ${APP} --home /app \
+ && echo -e "#!/bin/sh\nexec /usr/local/bin/${APP} \$@" > /app/entrypoint.sh \
```
fatedier, line 13: What's the purpose of this script? What are the differences compared to the previous execution method?

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+WORKDIR /building
+COPY . .
+
+ARG APP
+RUN make ${APP}
+
+FROM alpine:3.18 AS runtime
+
+ARG APP
+RUN addgroup -g 1000 -S ${APP} && adduser -u 1000 -S ${APP} -G ${APP} --home /app \
```
fatedier, line 12: Given that frp is often used to listen on port 80, will not using the root user have any impact? On the other hand, what problems may arise from using the root user?

fatedier:
I am not very professional in many aspects.

From my perspective, I am currently replacing some very simple and easy-to-understand Dockfile with a more complex version. I feel that there needs to be sufficient reasons to explain why we must do this.

I don't want to frequently modify these configurations/scripts that are unrelated to the core code unless there is official documentation stating that certain methods are explicitly prohibited or not recommended. This is similar to submitting a PR to modify code style, which may sometimes be acceptable but for maintainers, the benefits are minimal and it only increases maintenance costs.

Dockerfile
reneleonhardt: It' best practice to use specific versions, and it's effectively the same version (at the moment).
So if a project decides to improve security by rebuilding images even without code change, then version numbers make it easier for humans to accept the suggestions dependabot makes.
https://github.com/hadolint/hadolint/wiki/DL3007

Dockerfile
reneleonhardt: Best practice, a docker image tag is a also only a "label", those labels make it easier for your users what's inside this image.
The specific app version would be good too, but a bit harder to integrate in this case.
https://snyk.io/blog/how-and-when-to-use-docker-labels-oci-container-annotations/

Dockerfile
reneleonhardt: The reason for introducing this script was the unification of the 2 Dockerfiles into 1.
It's not allowed to use ARG/ENV variables inside ENTRYPOINT, your binaries have different names, it's best practice to create a minimal `exec` script in this case.
I will improve it a bit to make hadolint happy 😄

Dockerfile
reneleonhardt: Using root opens the attack surface to a whole lot more security bug possibilities.
The privileged port problem was solved 3 years ago:
https://stackoverflow.com/questions/66431299/i-can-bind-to-port-80-as-a-non-root-user-in-a-docker-container-why-whats-goin 
Can you try both new images to verify the expected functionality?

reneleonhardt:
I wasn't accusing you of anything, I am a professional 😄

You don't have to accept it if you don't want it.
I just improved all aspects including the content of the actual Dockerfiles.
For me it's 2 times easier to maintain 1 Dockerfile instead of 2 Dockerfiles doing "mostly" the same 😉

BustedSec:
I think the addition of not running the code as root is a very good improvement. I think this is a good request and the guy did most of the heavy lifting already. Not seeing any issues with his changes. Would recommend pushing change.

Dockerfile
fatedier: I do not prefer to add these redundant information.

fatedier:
The build time of the new Dockerfile increased from 43s to 188s on my mac.

Perhaps some changes are indeed necessary, but I find it difficult to accept the complexity introduced into originally simple rules. I don't see where the core value and benefits lie.

Maybe we can just make the necessary changes.

wuai1024:
我试过，如果不以root权限启动，就不能添加1024以下的端口了。

PR #3757, opened 2023-11-08
feat: allow dumping of configuration JSON Schema

akx:
### <samp>🤖 Generated by Copilot at 657e592</samp>

Add a feature to dump the JSON schema of the client and server configuration to stdout. This feature uses the `reflect` and `util` packages and introduces new flags to the `frpc` and `frps` commands. It also adds new dependencies to the `go.mod` file.

### WHY

For software that needs to generate frp configuration programmatically, it's useful to be able to validate the generated configuration before passing it to frp.

That software can now do `./frpc --json_schema > frpc.config.json` and use their favorite JSON Schema validator from there on out.

For instance, the generated client configuration schema looks like this: https://gist.github.com/akx/5c1a605e2624ee75131b4a684f235a82

Powered by https://github.com/invopop/jsonschema (❤️)

### DIFF
--- a/cmd/frpc/sub/root.go
+++ b/cmd/frpc/sub/root.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"reflect"
 	"sync"
 	"syscall"
 	"time"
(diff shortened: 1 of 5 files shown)


fatedier:
This seems like a capability that can be achieved outside of the core project, so I'm not too keen on including it in the command line parameters.

In addition, we already have a `verify` command in frpc, so perhaps it would be more appropriate to optimize it there.

akx:
@fatedier Thanks for the input! If including this capability in the shipped binaries is not OK: I'm not an experienced gopher, so can you recommend a good way to include this in a reproducible way (e.g. so that the schema can be shipped with each new version of frp, or whatnot)?

akx:
Or, to put it the other way: what is the best way to get a JSON Schema for the configuration format, if this isn't OK :)

fatedier:
What are your actual needs? If you need to verify if the configuration file is correct, we provide the "verify" command.

Additionally, if the current approach is to generate JSON Schema without modifying the core code, could this be another independent tool?

In the expected future, we will be conducting a series of refactoring. I am not very keen on introducing such non-core features at this early stage as they may be disrupted later, which could result in a worse user experience.

fatedier:
By the way, it is also possible to point out whether there are or how to implement similar features in some well-known projects, if it is possible to comprehensively learn from mature designs, that would be ideal.

akx:
My actual UC is that I have software that generates `frpc` configuration, and I'd like to add config validation to its test suite, so I could be reasonably confident that the generated configuration will work when the software is actually deployed.

Now that I actually got around to implement that using a schema generated from this PR, it turns out the generated schema is somewhat incorrect; e.g. `version` at the top level appears to be required, as does `log.maxDays`...

Maybe I'll just defer to shelling out to `frpc verify ...` instead. :)

fatedier:
#3773

PR #3848, opened 2023-12-15
fixed: API error when the proxy name contains "#".

mrasong:
### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at 4991596</samp>

This pull request fixes a bug that prevented proxies with special characters in their names from displaying traffic statistics in the web dashboard. It encodes the proxy name in the URL query string on the client side and decodes it on the server side in the dashboard API.

### WHY

fixed: API error when the proxy name contains "#".

### DIFF
--- a/server/dashboard_api.go
+++ b/server/dashboard_api.go
@@ -17,6 +17,7 @@ package server
 import (
 	"encoding/json"
 	"net/http"
+	"net/url"
 
 	"github.com/gorilla/mux"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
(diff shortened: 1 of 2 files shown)


fatedier:
I tend to prefer adding validation to disallow the use of special characters in the name. This will help avoid many unexpected issues.

mrasong:
> I tend to prefer adding validation to disallow the use of special characters in the name. This will help avoid many unexpected issues.

Although the validation that is added, you should also filter the URI which users use as a name.

PR #3871, opened 2023-12-21
fix(backoff): close of closed out channel

wuqinqiang:
### WHY

currently, MergeAndCloseOnAnyStopChannel may repeatedly call close out chan multiple times. even if Try0 recovered the panic. we should make sure that only once

### DIFF
--- a/pkg/util/wait/backoff.go
+++ b/pkg/util/wait/backoff.go
@@ -19,8 +19,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/samber/lo"
-
 	"github.com/fatedier/frp/pkg/util/util"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/util/wait/backoff.go
```diff
@@ -182,13 +183,15 @@ func Until(f func(), period time.Duration, stopCh <-chan struct{}) {
 
 func MergeAndCloseOnAnyStopChannel[T any](upstreams ...<-chan T) <-chan T {
 	out := make(chan T)
-
+	closeOnce := sync.Once{}
 	for _, upstream := range upstreams {
 		ch := upstream
 		go lo.Try0(func() {
```
fatedier, line 189: `Try0` is not needed?

wuqinqiang: > `Try0` is not needed if not repeating "close" anymore?
yes, done

fatedier (approved):

PR #3873, opened 2023-12-21
feat(backoff): add  BackoffUntilNil func in backoff

wuqinqiang:
### WHY

When loopLoginUntilSuccess is executed, it exits on success without using a channel. So add a new BackoffUntilNil function to handle this.
@https://github.com/fatedier/frp/issues/3860


fatedier:
We should not be inclined to add repetitive code. Instead, we should first try to see if there is a more elegant solution.

wuqinqiang:
> We should not be inclined to add repetitive code. Instead, we should first try to see if there is a more elegant solution.



Adjust BackoffUntil, make sliding, control exit, and other configurations optional, and decide whether to exit at nil based on the configuration parameters. What do you think?

PR #3900, opened 2023-12-28
fix: fix frpc default config filename

FarmerChillax:
### WHY

The configuration file cannot be found when starting with `./frpc` by default.

### DIFF
--- a/cmd/frpc/sub/root.go
+++ b/cmd/frpc/sub/root.go
@@ -43,7 +43,7 @@ var (
 )
 
 func init() {
-	rootCmd.PersistentFlags().StringVarP(&cfgFile, "config", "c", "./frpc.ini", "config file of frpc")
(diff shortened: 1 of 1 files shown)


fatedier:
This is a destructive change that may cause the application to fail to start properly in some users' environments. We will not make such changes in the short term, and it is also not recommended to use the configuration file path without specifying it explicitly through parameters.

FarmerChillax:
This modification is mainly to solve the current inconsistency between the default parameters and the documentation.

> This is a destructive change that may cause the application to fail to start properly in some users' environments.

If users actively update, there is bound to be this risk. They can still use the original configuration file via the `-c` parameter

fatedier:
Please specify the configuration file according to the document. The default value may be deprecated in future versions.

PR #4020, opened 2024-02-22
use math/rand/v2

fatedier:


fatedier:
Wait the new golangci-lint version

fatedier:
https://github.com/golangci/golangci-lint/issues/4415

PR #4084, opened 2024-03-19
add header for http healthcheck

KaiveYoung:

### DIFF
--- a/client/health/health.go
+++ b/client/health/health.go
@@ -40,8 +40,8 @@ type Monitor struct {
 	addr string
 
 	// For http
-	url string
-
+	url            string
+	header         http.Header
 	failedTimes    uint64
 	statusOK       bool
 	statusNormalFn func()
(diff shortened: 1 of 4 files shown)


fatedier:
LGTM for adding this feature.

But I do not want to reuse the HeaderOperations structure. There may be add/remove operations in HeaderOperations in the future, which do not fully match semantically.

For this part, I tend to refer to the field design in K8s: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-http-request

```go
// HTTPHeader describes a custom header to be used in HTTP probes
type HTTPHeader struct {
    // The header field name
    Name string
    // The header field value
    Value string
}

// HTTPGetAction describes an action based on HTTP Get requests.
type HTTPGetAction struct {
    // Optional: Path to access on the HTTP server.
    // +optional
    Path string
    // Required: Name or number of the port to access on the container.
    // +optional
    Port intstr.IntOrString
    // Optional: Host name to connect to, defaults to the pod IP. You
    // probably want to set "Host" in httpHeaders instead.
    // +optional
    Host string
    // Optional: Scheme to use for connecting to the host, defaults to HTTP.
    // +optional
    Scheme URIScheme
    // Optional: Custom headers to set in the request. HTTP allows repeated headers.
    // +optional
    HTTPHeaders []HTTPHeader
}
```

The important thing to note is that "Host" in Go may be a unique header, typically configured using a separate field.

KaiveYoung:
Okay, I'll rewrite it. And I also forgot to delete the debug code.

PR #4085, opened 2024-03-19
add header for http healthcheck

KaiveYoung:

### DIFF
--- a/client/health/health.go
+++ b/client/health/health.go
@@ -74,7 +74,7 @@ func NewMonitor(ctx context.Context, cfg v1.HealthCheckConfig, addr string,
 		url = s + cfg.Path
 	}
 	header := make(http.Header)
-	for _, h := range cfg.Headers {
+	for _, h := range cfg.HTTPHeaders {
 		header.Set(h.Name, h.Value)
 	}
 
--- a/client/health/health.go
+++ b/client/health/health.go
@@ -40,8 +40,9 @@ type Monitor struct {
 	addr string
 
 	// For http
-	url string
-
+	url            string
+	header         http.Header
+	host           string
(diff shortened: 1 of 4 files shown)


client/health/health.go
```diff
@@ -163,6 +171,8 @@ func (monitor *Monitor) doHTTPCheck(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+	req.Header = monitor.header
+	req.Host = monitor.host
```
fatedier, line 175: I have some misunderstandings about the configuration of the host in K8s. We should not add the host parameter here, but instead actively extract the host field from the header and assign it to req.Host.

```go
 req.Host = monitor.header.Get("Host")
```

fatedier (changes requested):

fatedier:
By the way, you don't need to create a new PR, just push the new changes to the original branch.

pkg/config/v1/proxy.go
```diff
@@ -97,6 +97,12 @@ type HealthCheckConfig struct {
 	// Path specifies the path to send health checks to if the
 	// health check type is "http".
 	Path string `json:"path,omitempty"`
+	// Headers specifies the headers to send health checks to if the
+	// health check type is "http".
+	Headers []HTTPHeader `json:"headers,omitempty"`
```
fatedier, line 102:
```suggestion
	HTTPHeaders []HTTPHeader `json:"httpHeaders,omitempty"`
```

KaiveYoung:
> By the way, you don't need to create a new PR, just push the new changes to the original branch.

New to PR, don’t know how to use it

fatedier (approved):

PR #4107, opened 2024-03-27
Preferred Toml

csznet:
### WHY

下载的包默认为toml的sample，但是frpc默认的还是frpc.ini，将默认文件改为frpc.toml


fatedier:
This is the expected behavior in order to maintain compatibility.

Also, please do not use the default configuration file path, as this often leads to various issues. We may remove this default value in future versions.

PR #4153, opened 2024-04-11
fix: ssh tunnel with user name will lead to invalid proxy name

blizard863:


pkg/ssh/server.go
```diff
@@ -109,7 +109,6 @@ func (s *TunnelServer) Run() error {
 	if sshConn.Permissions != nil {
 		clientCfg.User = util.EmptyOr(sshConn.Permissions.Extensions["user"], clientCfg.User)
 	}
-	pc.Complete(clientCfg.User)
```
fatedier, line 112: This is as expected.

PR #4161, opened 2024-04-16
Add Helm charts for FRP server and client

koendelaat:
### WHY

For easy deployment on kubernetes two helm charts are created.

This allows forwarding kubernetes services using FRP


koendelaat:
Ok, I'll keep it in a repo under my own account.

PR #4191, opened 2024-04-28
initialize tls config with TLS 1.2

yuyoyuppe:
### WHY

Fix frpc issue that prevents obtaining A-grade SSL rating from [ssl labs](https://www.ssllabs.com/) caused by incorrect crypto/tls initialization that supports disabled TLS versions.

### DIFF
--- a/pkg/plugin/client/https2http.go
+++ b/pkg/plugin/client/https2http.go
@@ -104,7 +104,8 @@ func (p *HTTPS2HTTPPlugin) genTLSConfig() (*tls.Config, error) {
 		return nil, err
 	}
 
-	config := &tls.Config{Certificates: []tls.Certificate{cert}}
+	config := &tls.Config{Certificates: []tls.Certificate{cert}, MinVersion: tls.VersionTLS12}
+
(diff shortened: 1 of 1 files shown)


fatedier:
This is the default behavior of Go 1.22, no modifications are needed.

yuyoyuppe:
frpc v0.57.0 on Windows still has this issue. Is this just a pipeline thing then that would be fixed in the next release?

fatedier:
I suggest you to write a simple Go application in your own environment and use the Go 1.22 compiler to further investigate possible reasons for the issue.

We are currently not inclined to make any changes in this area, but instead use the official default values, which are usually the result of various factors considered by the go community, such as perhaps taking into account widespread compatibility issues.

The ability for more advanced configurations is planned for v2, but will not be supported in the short term.

yuyoyuppe:
You're right, I've just done a further testing and the problem is indeed fixed by just compiling frpc from the `dev` branch with the latest Go compiler. 

So it's just a pipeline issue indeed.

Closing, thanks!

PR #4256, opened 2024-06-01
feat: add client reload on config change

obadoraibu:
Thid pull request proposed as part of the homework for the course "designing the architecture of software systems". The ability to configure client updates when configuration files are changed is implemented. "Watcher" was used instead of "fsnotify" due to problems with the compatibility of the second with macos


fatedier:
https://github.com/fatedier/frp/pull/3728#issuecomment-1789992965

This feature is not currently in the plan due to some concerns and it is not very perfect. For now, you can reload the configuration through the reload API after updating the configuration file.

PR #4311, opened 2024-06-30
feat: use sprig (slimmed down) for template function map

debdutdeb:
### WHY

Gives more control over to how we can set our configs.

For example since not everything has an flag cmdline, one could simplify the interface through environment variables instead of editing the config file, manually or automatically.

e.g.

```toml
{{ $hosts := splitList "," .Envs.HOSTS }}
{{ range $index, $port := splitList "," .Envs.PORTS }}
[[proxies]]
name = {{ printf "web%d" $index | trim | quote }}
type = "http"
localPort = {{ $port }}
customDomains = [ {{ printf "%s.tunnel.example.com" (index $hosts $index) | trim | quote }} ]
{{ end }}
```

So I can have an function to simplify the interface like

```sh
# use like `frpc 8080 echoserver` should make localhost:8080 accessible at echoserver.tunnel.example.com
frpc() {
	local \
		port="$1" \
		name="$2"

	export HOSTS="$name"
	export PORTS="$port"

	command frpc -c ~/.config/frpc/config.toml
}
```

### Note

Using `slim-sprig` because of 

> with all functions that depend on external (non standard library) or crypto packages removed

As noted in https://github.com/go-task/slim-sprig

Long term seems like there is an attempt at forking and maintaining separate version - https://github.com/go-sprout/sprout

I can change to sprout as is, but decided to start off with a lightweight version instead. Since still pre major 1, there's still the freedom to experiment.

Let me know whichever you prefer :)


fatedier:
This is not a project focused on configuration and deployment... I am not inclined to maintain this aspect of capability.

PR #4329, opened 2024-07-10
fix misuse of unlock

fakeboboliu:
### WHY

phase, which needs protection, is still out of critical section


client/proxy/proxy_wrapper.go
```diff
@@ -254,8 +254,8 @@ func (pw *Wrapper) statusFailedCallback() {
 func (pw *Wrapper) InWorkConn(workConn net.Conn, m *msg.StartWorkConn) {
 	xl := pw.xl
 	pw.mu.RLock()
+	defer pw.mu.RUnlock()
```
fatedier, line 257: `phase := pw.Phase`

You can do it like this, without increasing the scope of the lock.

PR #4354, opened 2024-07-29
Fix loginFailExit = false bug

Yurunsoft:
### WHY

Fixed the issue that when loginFailExit = false, the frpc stop command cannot be stopped correctly if the server is not successfully connected after startup


fatedier:
Can you please add a release note to https://github.com/fatedier/frp/blob/dev/Release.md? I believe this qualifies as a fix.

Yurunsoft:
> Can you please add a release note to https://github.com/fatedier/frp/blob/dev/Release.md? I believe this qualifies as a fix.

Done.

PR #4422, opened 2024-09-06
Feature: Add remove api for offline proxy

peter-wangxu:
This PR adds api to remove a single proxy with name and type example:
  curl -X DELETE -u admin:admin -v http://127.0.0.1:7500/api/proxy/tcp/test-tcp
output:
  proxy "test-tcp" deleted

### WHY

Currently, only clear all offline proxy api is provided, this PR provides api for singe proxy removal

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 # Folders
 _obj
 _test
+.DS_Store
 
 # Architecture specific extensions/prefixes
 *.[568vq]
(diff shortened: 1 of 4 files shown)


peter-wangxu:
@fatedier Please take a review

fatedier:
Currently, there are no plans to support this API, and only the option to clear all is provided in the dashboard.

peter-wangxu:
> Currently, there are no plans to support this API, and only the option to clear all is provided in the dashboard.

I am adding via this PR.

In our case, we check the proxy status via get api, we also want delete api to manually remove some unwanted or unexpected proxies on purpose.

PR #4431, opened 2024-09-12
fix: memory leak

ivanszl:
### WHY
在某些特殊情况下，会导致内存泄露
在发送比接收慢的情况下，容易出现sendCh满，会导致后续所有的 goroutine 泄露了

### DIFF
--- a/client/control.go
+++ b/client/control.go
@@ -97,7 +97,7 @@ func NewControl(ctx context.Context, sessionCtx *SessionContext) (*Control, erro
 		ctl.msgDispatcher = msg.NewDispatcher(sessionCtx.Conn)
 	}
 	ctl.registerMsgHandlers()
-	ctl.msgTransporter = transport.NewMessageTransporter(ctl.msgDispatcher.SendChannel())
+	ctl.msgTransporter = transport.NewMessageTransporter(ctl.msgDispatcher.Send)
 
 	ctl.pm = proxy.NewManager(ctl.ctx, sessionCtx.Common, ctl.msgTransporter)
 	ctl.vm = visitor.NewManager(ctl.ctx, sessionCtx.RunID, sessionCtx.Common, ctl.connectServer, ctl.msgTransporter)
--- a/pkg/msg/handler.go
+++ b/pkg/msg/handler.go
@@ -54,6 +54,10 @@ func (d *Dispatcher) sendLoop() {
 	for {
 		select {
 		case <-d.doneCh:
+			// need to clear all message
+			// Otherwise, it will memory leak when sendCh is full.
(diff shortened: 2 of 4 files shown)


fatedier (commented):
Make only necessary modifications and avoid changing additional areas according to your own habits.

Channels will be garbage collected after they are no longer referenced, so there is no need to explicitly read all elements.

Try to provide reproducible steps and scenarios.

pkg/transport/message.go
```diff
@@ -54,7 +54,7 @@ type transporterImpl struct {
 
 func (impl *transporterImpl) Send(m msg.Message) error {
```
fatedier, line 55: We intend to add a context parameter similar to the Do function to allow cancellation of sending.

One case is when the user exits manually, which can be canceled through this mechanism, and another case is for potential scenarios where a timeout might be set later.

fatedier, line 55: Additionally, a TrySend function can be added, which returns a specific error when sendCh is full, instead of blocking. This function is more suitable for certain specific scenarios.

pkg/msg/handler.go
```diff
@@ -66,6 +66,7 @@ func (d *Dispatcher) readLoop() {
 		m, err := ReadMsg(d.rw)
 		if err != nil {
 			close(d.doneCh)
+			close(d.sendCh)
```
fatedier, line 69: sendCh does not need to be actively closed; it should be automatically reclaimed after all senders have exited.

As I commented in transport/message.go, I prefer to use a reasonable mechanism to ensure exit on the sender's side.

client/control.go
```diff
@@ -97,7 +97,7 @@ func NewControl(ctx context.Context, sessionCtx *SessionContext) (*Control, erro
 		ctl.msgDispatcher = msg.NewDispatcher(sessionCtx.Conn)
 	}
 	ctl.registerMsgHandlers()
-	ctl.msgTransporter = transport.NewMessageTransporter(ctl.msgDispatcher.SendChannel())
+	ctl.msgTransporter = transport.NewMessageTransporter(ctl.msgDispatcher.Send)
```
fatedier, line 100: I hope to keep it as it is.

pkg/transport/message.go
ivanszl: Adding context parameters would be a better solution.
In other cases, adding a timeout is indeed a viable solution, and I have considered using this pattern before

ivanszl: Yes，Adding TrySend is indeed a way to make it clearer. I didn't read the code thoroughly and may not fully understand the this function.

PR #4526, opened 2024-11-06
Add the latest version of the image.

wuai1024:
Add a latest TAB to the original image version to mark the latest version. Easy image pulling with fatedier/frp.

Fix [#4525](https://github.com/fatedier/frp/issues/4525)


zzzshanghai (commented):
Thank you very much.

PR #4561, opened 2024-11-28
config: add VhostEnableH2C flag to support HTTP/2 cleartext upgrade

### LINKED ISSUE #4563: [Feature Request] Support h2c in vhost server

Like the title, we can enable `h2c` support in vhost server, there is a PR already

https://github.com/fatedier/frp/pull/4561

_No response_

### Affected area

- [ ] Docs
- [ ] Installation
- [ ] Security
- [ ] User Experience
- [ ] Test and Release
- [ ] Client Plugin
- [ ] Server Plugin
- [ ] Extensions
- [ ] Others

sword-jin:
### WHY

Support `h2c` in virtual host.

Our use case is:

In front of the vhost, we deploy a caddy to reverse the traffic to frps. And use caddy do tls termination, we got a bug, the client always get an error `internal: protocol error: no Grpc-Status trailer: unexpected EOF` when requesting a GRPC streaming API(works well on unary API).

After some investigation and debug, I found we can fix it by support h2c on the vhost layer.

refrence: 

- Issue: https://github.com/golang/go/issues/14141

close #4563


fatedier:
The better approach is to submit an issue first, engage in a discussion, reach a consensus, and then submit the code.

fatedier:
I'd prefer not to introduce new optional parameters.  Maybe we can support this functionality transparently so users don't need to make any changes.

PR #4577, opened 2024-12-09
Upgrade packages to resolve CVE-2024-53259

AdallomRoy:
### WHY

Upgrading packages to resolve CVE-2024-53259


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/fatedier/frp
 
-go 1.22
+go 1.22.0
+
+toolchain go1.23.4
```
fatedier, line 5: Can we skip adding this?

AdallomRoy: I tried, but the linter failed claiming that go mod tidy should run.
You can try removing that with go 1.22 and running go mod tidy to see what happens, it might provide better results.

fatedier: I did not encounter any issues in my local go1.23.3 environment.

AdallomRoy: See now, without that mod tidy

fatedier: Strange problem, it seems that `go 1.22.0` needs to be retained.

AdallomRoy: Done

fatedier (approved):

PR #4606, opened 2024-12-24
Implement HTTP/2 to HTTP/1.1 downgrade support if the request is GRPC…

sword-jin:
resolve: #4563 

force to use h2 transport only for grpc request


fatedier:
This implementation does not look elegant and is not a general-purpose solution. I am not inclined to introduce such logic.

sword-jin:
> This implementation does not look elegant and is not a general-purpose solution. I am not inclined to introduce such logic.

could you provide another acceptable idea?

PR #4628, opened 2025-01-08
Added Helm chart for frps.

ckt114:
### WHY

Ability to run frps in Kubernetes.


fatedier:
https://github.com/fatedier/frp/pull/4161#issuecomment-2061150387

ckt114:
Okay that's fine. Other apps have helm deployment in the same repo that's why I thought it would be good to do it here.

PR #4638, opened 2025-01-15
解决goroutine泄漏问题

Alpha2J:
### WHY
由于`closeCh`没有被初始化，在`Close()`函数中执行`close(pxy.closeCh)`的时候，会panic："panic error: close of nil channel"；导致无法退出`Run()`函数中创建的goroutine，造成goroutine泄漏，同时引发内存泄漏；在我的案例中，client用户单实例大概4000，运行两天，阻塞在`case <-pxy.closeCh`中的goroutine达到三万多；

### DIFF
--- a/server/proxy/xtcp.go
+++ b/server/proxy/xtcp.go
@@ -17,6 +17,7 @@ package proxy
 import (
 	"fmt"
 	"reflect"
+	"sync"
 
 	"github.com/fatedier/golib/errors"
 
@@ -32,7 +33,8 @@ type XTCPProxy struct {
 	*BaseProxy
 	cfg *v1.XTCPProxyConfig
 
-	closeCh chan struct{}
+	closeCh   chan struct{}
+	closeOnce sync.Once
 }
 
 func NewXTCPProxy(baseProxy *BaseProxy) Proxy {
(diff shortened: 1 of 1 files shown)


fatedier:
```go
func (pxy *XTCPProxy) Close() {
	pxy.BaseProxy.Close()
	pxy.rc.NatHoleController.CloseClient(pxy.GetName())
	_ = errors.PanicToError(func() {
		close(pxy.closeCh)
	})
}
```

Could you also update this to use `sync.Once` to ensure it only closes once? This way, we can avoid using `PanicToError`.

Alpha2J:
> Could you also update this to use `sync.Once` to ensure it only closes once? This way, we can avoid using `PanicToError`.

Sure, I've updated the code to use `sync.Once` to ensure that `close(pxy.closeCh)` is only called once. However, I'm still not quite sure why `close(pxy.closeCh)`needs to be wrapped with `PanicToError`. Could you please clarify the reasoning behind that? Is it due to some historical reasons or edge cases that need to be handled?

fatedier:
> This way, we can avoid using PanicToError.
Just remove it.

server/proxy/xtcp.go
```diff
@@ -89,7 +92,9 @@ func (pxy *XTCPProxy) Run() (remoteAddr string, err error) {
 func (pxy *XTCPProxy) Close() {
 	pxy.BaseProxy.Close()
 	pxy.rc.NatHoleController.CloseClient(pxy.GetName())
-	_ = errors.PanicToError(func() {
-		close(pxy.closeCh)
+	pxy.closeOnce.Do(func() {
```
fatedier, line 95: The two function calls above can also be included.

fatedier (approved):

PR #4733, opened 2025-03-29
add callback option to client service to track login state

koho:
### WHY

When using frp as library, I want to check if the connection between frpc and frps is working properly.

In most cases, we can determine that by the proxy status, for example, if the login was unsuccessful, the API `/api/status` will respond with an empty object. However, this method is not working if no proxy is provided in the config.

It would be great if a callback function could be added to client service to allow users to track the login state. I'm not sure this single callback function is sufficient to cover common use cases. If we should do more work, I think we can add more callbacks, such as `on_connect(msg.LoginResp)`, `on_connect_fail` and `on_disconnect`. In my opinion, since we currently don't have error codes defined in login response, the callback function `on_connect(msg.LoginResp)` may be replaced with `on_connect(bool)`. As for the callback `on_connect_fail`, it may not be commonly used.

### DIFF
--- a/client/service.go
+++ b/client/service.go
@@ -84,6 +84,9 @@ type ServiceOptions struct {
 	//
 	// If it is not set, the default frpc implementation will be used.
 	HandleWorkConnCb func(*v1.ProxyBaseConfig, net.Conn, *msg.StartWorkConn) bool
+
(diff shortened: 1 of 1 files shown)


lingchenma:
Extend life

PR #4827, opened 2025-06-05
vnet: initial Windows support

EZForever:
### WHY

VirtualNet is a very promising feature, and the underlying TUN library (`golang.zx2c4.com/wireguard/tun`) already supports Windows, so... why not ;)

### Additional Considerations

`golang.zx2c4.com/wireguard/tun` on Windows relies on [wintun](https://www.wintun.net/) to do its job, namely the `wintun.dll` file from the prebuilt binaries package. Including it as a built-in dependency is out of question, since wintun is in C, and we need the (signed) prebuilt driver binaries anyway for dealing with Windows' driver signing enforcement. Maybe we could bundle it with Windows release builds, but that kinda breaks the zero-dependency promise. It is not a concerning problem right now since VirtualNet is still in Alpha stage and users are assumed to know what they're doing; we could just direct them to manually download the required binaries in the docs.

P.S. This is my first time coding in Go, so please forgive me for stupid mistakes in code.

### DIFF
--- a/pkg/vnet/tun_unsupported.go
+++ b/pkg/vnet/tun_unsupported.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build !darwin && !linux
+//go:build !darwin && !linux && !windows
 
 package vnet
 
--- a/pkg/vnet/tun_windows.go
+++ b/pkg/vnet/tun_windows.go
@@ -1,0 +1,124 @@
+// Copyright 2025 The frp Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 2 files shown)


fatedier:
The functionality on Windows depends on external components, which is why I didn’t consider supporting Windows right away. I intend to ensure the core features are stable on macOS and Linux first, and only then think about how to support other systems more elegantly.

The code itself isn't that important; testing across various environments and scenarios, as well as the related documentation, matters more.

Could you try doing more testing and validation, and organizing the related documentation to help move this forward?

EZForever:
I do not know enough Go and/or software engineering to conduct robust testing on many environments and scenarios. Since the code here is not important, I'll just draft this PR for now.

Issue #1788, opened 2020-05-01
[feature request] Add client count metrics broken down by name in prometheus metrics


Lilja:
Currently, it is not possible to see which clients are online in a dashboard.

There is currently this:
https://github.com/fatedier/frp/blob/23bb76397a0484485a16e1a3e990385b13ee388d/models/metrics/prometheus/server.go#L58-L63

Which is an integer of the total amount of clients connected to frps. But I wish that there could be another metric `clients` that shows the total amount of clients by name. So if I have a service/client like this:
```
# frpc.ini
[my-cool-client]
...
```

Then I would see in `ip_address:7500/metrics`:

> \# HELP frp_server_client_counts The current client counts of frps
> \# TYPE frp_server_client_counts gauge
> frp_server_client_counts 1
> **\# HELP frp_server_clients The current clients of frps broken down by name
> \# TYPE frp_server_clients gauge
> frp_server_clients{name="my-cool-service"} 1**
> \# HELP frp_server_connection_counts The current connection counts
> \# TYPE frp_server_connection_counts gauge
> frp_server_connection_counts{name="my-cool-service",type="tcp"} 1
> frp_server_connection_counts{name="ssh",type="tcp"} 0

Note the **bolded**, which are what I would like to achive.

TL;DR: I'm trying to replicate the dashboard page in grafana and I would like to know which clients are online, not how many clients are online.

![image](https://user-images.githubusercontent.com/6134511/80794852-34306500-8b9b-11ea-8c15-1f184612da2e.png)

but the prometheus `/metrics` page doesn't.

I would love to help by submitting a PR but I have never worked with `golang` so I'm afraid I wont be with too much help.

PR #4864, opened 2025-07-02
feat: add proxy_status metric

MelvinHuls:
Added a proxy_status metric that is either 1 (online) or 0 (offline).

### WHY

We want to be able to tell when a customers proxy goes offline, and do alerting on it.

### DIFF
--- a/pkg/metrics/aggregate/server.go
+++ b/pkg/metrics/aggregate/server.go
@@ -68,6 +68,12 @@ func (m *serverMetrics) CloseProxy(name string, proxyType string) {
 	}
 }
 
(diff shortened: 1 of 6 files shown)


MelvinHuls:
Saw that for contribution an issue should be created first, but there is already one open which is very similar to what I want to achieve: https://github.com/fatedier/frp/issues/1788

MelvinHuls:
What in the modern tamagotchi? Don't go and expire on me now.
Is a comment enough to keep this from being marked as rotten? Or do I need to invest into a virtual refrigerator?

MelvinHuls:
So that's it? I go on holiday for 2 weeks and my Neopet dies?
Never though I would get to the point of needing to hire a virtual pet sitter.

PR #4891, opened 2025-07-21
Add `install:` to Makefile

donmor:
Add `install:` target that can be invoked by `# make install`; DESTDIR/prefix is supported

### WHY

For better packaging


fatedier:
I don't think we need this command.

Issue #4972, opened 2025-09-08
[Feature Request] Custom response for unregistered HTTP vhost (rules, wildcard, headers)


mnasruul:
## Summary
Add support for configurable `customResponse` in FRPS HTTP vhost, so requests with unknown/unregistered Host header return a user-defined response (status, content-type, body, headers).

## Why
Currently, FRPS always returns default 404 for unregistered hostnames. This is not customizable. Custom responses improve UX and debugging, especially in multi-tenant setups.

```
## Proposed config (TOML)
[customResponse]
enable = false

[[customResponse.rules]]
hostname    = ["*.example.com", "example.com"]
statusCode  = 503
contentType = "text/html"
body = """
<!doctype html><html><body>
<h1>Service Unavailable</h1>
<p>The server is currently unavailable.<br/>Please try again later.</p>
</body></html>
"""
[customResponse.rules.headers]
Cache-Control = "no-store"
X-Error-Code  = "UNREGISTERED_HOST"

[[customResponse.rules]]
hostname    = ["spesific.example2.com"]
statusCode  = 404
contentType = "application/json"
body        = "{\"error\":\"unregistered_host\",\"hint\":\"register frpc for this hostname\"}"
[customResponse.rules.headers]
Cache-Control = "no-store"
```

## Behavior
- If feature disabled, behavior unchanged (default 404).
- Rules support exact, wildcard (`*.example.com`), and `"*"` catch-all.
- First match wins.

## Next step
I can prepare a PR with implementation + unit tests if this direction is acceptable.

### Affected area

- [x] Docs
- [ ] Installation
- [ ] Performance and Scalability
- [ ] Security
- [x] User Experience
- [ ] Test and Release
- [ ] Developer Infrastructure
- [ ] Client Plugin
- [x] Server Plugin
- [ ] Extensions
- [ ] Others

fatedier:
Thanks for the proposal! We’re not expanding L7 HTTP/HTTPS features right now. The current L7 pieces are fragmented and need a redesign. We’ll revisit ideas like this after the L7 refactor. For now, please consider putting a reverse proxy (Nginx/Caddy/Traefik) in front of frps to handle custom fallback pages. Appreciate your understanding!

mnasruul:
As mentioned in the related PR, this proposal is on hold until the L7 HTTP/HTTPS refactor.

If anyone needs this feature earlier, I’ve published my fork with a working implementation here:
👉 https://github.com/mnasruul/frp

[...]

PR #4973, opened 2025-09-08
feat(vhost): add customResponse config for unregistered HTTP hostnames

mnasruul:
### **Summary**
This PR introduces a new customResponse configuration block in FRPS.
When a client sends an HTTP request with a Host that is not registered by any FRPC, FRPS can now return a custom response instead of the default 404 page.

### **Why**
Currently, unregistered host requests always return the same default 404 response.
With customResponse, users can define:

- Status code (e.g., 404, 503)
- Content type (e.g., text/html, application/json)
- Custom body (multi-line supported)
- Additional headers (map of key:value)

### This improves:

- User experience: clearer responses for clients.
- Multi-tenant setups: per-domain fallback behavior.

Configuration Example (frps.toml)
```
[customResponse]
# Enable or disable the feature. Default = false (disabled).
enable = false

# Rule 1: Return a 503 HTML page for *.example.com and example.com
[[customResponse.rules]]
hostname    = ["*.example.com", "example.com"]
statusCode  = 503
contentType = "text/html"
body = """
<!doctype html>
<html>
<body>
<h1>Service Unavailable</h1>
<p>The server is currently unavailable.<br/>
Please try again later.</p>
</body>
</html>
"""
[customResponse.rules.headers]
Cache-Control = "no-store"
X-Error-Code  = "UNREGISTERED_HOST"

# Rule 2: Return a 404 JSON payload for spesific.example2.com
[[customResponse.rules]]
hostname    = ["spesific.example2.com"]
statusCode  = 404
contentType = "application/json"
body        = "{\"error\":\"unregistered_host\",\"hint\":\"register frpc for this hostname\"}"
[customResponse.rules.headers]
Cache-Control = "no-store"
```

### Notes:
Rules are matched in order; first match wins.

- *.example.com does not match example.com. To cover both, list both hostnames.
- "*" can be used as a catch-all rule.
- If enable = false or no rule matches, FRPS falls back to current default behavior.

### Design Overview

- New config types in pkg/config/v1/server.go: CustomResponseConfig and CustomResponseRule.
- Internal mapping to vhost.CustomErrorPage in server/service.go (avoids import cycles).
- New matcher helper MatchDomain in pkg/util/http/http.go.
- ErrorHandler in pkg/util/vhost/http.go updated to call CustomErrorResponse.

### Compatibility

- Default behavior unchanged when [customResponse] is disabled (or not defined).
- Fully backward-compatible.

### Test Plan

- Unit tests for:
- Exact host match (example.com)
- Wildcard (*.example.com)
- Catch-all ("*")
- Hostnames with :port (normalized)
- No match (falls back to default 404)

Manual curl tests against FRPS with/without enabled rules.
Related Issue
Closes https://github.com/fatedier/frp/issues/4972

### DIFF
--- a/conf/frps_full_example.toml
+++ b/conf/frps_full_example.toml
@@ -167,3 +167,60 @@ name = "port-manager"
 addr = "127.0.0.1:9001"
 path = "/handler"
 ops = ["NewProxy"]
+
+# ========================================================================
+# Custom response configuration (for HTTP vhost requests)
+# ========================================================================
+# This section allows FRPS to return a custom response when a client sends
+# an HTTP request with a Host header that is not registered by any FRPC.
+#
+# - Set enable = true to turn this feature on.
+# - Rules are matched in order; the first match wins.
+# - Supported hostname patterns:
+#     * Exact domain: "example.com"
+#     * Wildcard:     "*.example.com" (matches foo.example.com, bar.example.com, but NOT example.com itself)
+#     * Catch-all:    "*" (matches any host)
+# - Each rule can define:
+#     * statusCode   : HTTP status code (e.g. 404, 503)
+#     * contentType  : MIME type of the response (e.g. text/html, application/json)
+#     * body         : Response body (string, multi-line supported with """ ... """)
+#     * headers      : Extra headers (map of key:value)
+#
+# If no rule matches or enable = false, FRPS falls back to the default 404.
+# ========================================================================
+
+[customResponse]
+# Enable or disable the custom response feature. Default = false (disabled).
+enable = false
+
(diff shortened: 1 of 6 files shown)


mnasruul:
Closing as per maintainer feedback (see comment). Will revisit after L7 refactor.
For now I’ll be using a reverse proxy (e.g. Nginx) to handle custom fallback pages as suggested.

If other users are interested in this feature before the L7 refactor, you can find my fork/clone with the implementation here:
👉 https://github.com/mnasruul/frp

Once the L7 layer has been redesigned, I’d be happy to contribute again and help bring this feature upstream. Thanks!

PR #4979, opened 2025-09-15
resolve shellcheck warnings

TheKhanj:
I’m not sure which shells you want to support, but for now ShellCheck supports sh, bash, dash, and ksh, and it was complaining about these issues. It doesn’t support zsh, ash, csh, or others—but most of its complaints are pretty reasonable (though you could definitely argue about that x-prefixing thing 🤷).

It’s a minor thing, I know… but I wanted to familiarize myself with the project a bit—and since I was already using a couple of these files in my own projects (not sure if I even have the permission to do that 🫣), I figured I’d fix them here too :). Feel free to ignore the pull request if it doesn’t fit.

Check exit code directly with e.g. 'if ! mycmd;', not indirectly with $?. (shellcheck SC2181) 
https://www.shellcheck.net/wiki/SC2181

Use $(...) notation instead of legacy backticks `...`. (shellcheck SC2006)
https://www.shellcheck.net/wiki/SC2006

Avoid x-prefix in comparisons as it no longer serves a purpose. (shellcheck SC2268) https://www.shellcheck.net/wiki/SC2268

Double quote to prevent globbing and word splitting.
(shellcheck SC2086)
https://www.shellcheck.net/wiki/SC2086

Double quote to prevent globbing and word splitting.
(shellcheck SC2086)
https://www.shellcheck.net/wiki/SC2086

### DIFF
--- a/hack/run-e2e.sh
+++ b/hack/run-e2e.sh
@@ -10,7 +10,7 @@ if ! command -v ginkgo >/dev/null 2>&1; then
 fi
 
 debug=false
-if [ "x${DEBUG}" = "xtrue" ]; then
+if [ "${DEBUG:-false}" = "true" ]; then
     debug=true
 fi
 logLevel=debug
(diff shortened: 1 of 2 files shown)


fatedier:
If the current version works, I’d rather not make too many changes in this area—unless there’s a clear case showing it fails under certain conditions.

TheKhanj:
Makes sense :)

PR #4983, opened 2025-09-17
tcp ip address whitelist on frps server

pipethedev:
# Add IP Address Whitelist for Proxied Services

## WHY

This feature addresses the need to restrict access to proxied services based on client IP addresses. Currently, FRP allows any IP to connect to proxied services once they're exposed. This creates security concerns for users who want to limit access to specific IP ranges or trusted networks.

### Configuration Options:

**TOML Configuration:**
```toml
# frps.toml
bindPort = 7000
allowedAccessIPs = ["127.0.0.1", "192.168.1.0/24", "10.0.0.0/8"]
```

**Command Line:**
```bash
./frps --allowed-access-ips 127.0.0.1,192.168.1.0/24
```

## Breaking Changes
None - this is a backward compatible feature that defaults to allowing all IPs when not configured.

### DIFF
--- a/conf/frps.toml
+++ b/conf/frps.toml
@@ -1,1 +1,4 @@
 bindPort = 7000
+
+# Uncomment to restrict client connections to specific IP addresses
+# allowedClientIPs = ["127.0.0.1", "192.168.1.0/24"]
--- a/conf/frps.toml
+++ b/conf/frps.toml
@@ -1,4 +1,4 @@
 bindPort = 7000
 
-# Uncomment to restrict client connections to specific IP addresses
-# allowedClientIPs = ["127.0.0.1", "192.168.1.0/24"]
+# Uncomment to restrict access to proxied services to specific IP addresses
+# allowedAccessIPs = ["127.0.0.1", "192.168.1.0/24"]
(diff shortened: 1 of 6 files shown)


fatedier:
https://github.com/fatedier/frp?tab=readme-ov-file#contributing

Please try to have a thorough discussion in the issue first before implementing the code.

At present, you can leverage the Server Plugin mechanism to implement custom plugins, and enforce IP whitelisting/blacklisting by handling the NewUserConn hook. 

There are currently no plans for native support of this feature, since supporting all proxy types—not just TCP—would require significant changes and a redesign. This is an optimization we plan to address in the next major release.

pipethedev:
> https://github.com/fatedier/frp?tab=readme-ov-file#contributing
> 
> 
> 
> Please try to have a thorough discussion in the issue first before implementing the code.
> 
> 
> 
> At present, you can leverage the Server Plugin mechanism to implement custom plugins, and enforce IP whitelisting/blacklisting by handling the NewUserConn hook. 
> 
> 
> 
> There are currently no plans for native support of this feature, since supporting all proxy types—not just TCP—would require significant changes and a redesign. This is an optimization we plan to address in the next major release.

When would there be a next major release ?
And would the next release support host name verification for tls connection ?

fatedier:
https://github.com/fatedier/frp?tab=readme-ov-file#about-v2

There is currently no timeline for v2.

> And would the next release support host name verification for tls connection ?

I'm not sure what this refers to, could you reference the specific issue?

PR #4985, opened 2025-09-19
add proxy name label to the proxy_count prometheus metric

### LINKED ISSUE #4986: [Feature Request] in prometheus metrics, I would like to see which proxies are currently registered (i.e. frpc connected)

Since proxy registration seems to imply that the frpc client is online, it would be great to add the "name" label to this metric. I have made a [PR to add this here](https://github.com/fatedier/frp/pull/4985)

I was wondering if this was tried before and maybe an issue was encountered?

_No response_

### Affected area

- [ ] Docs
- [ ] Installation
- [ ] Security
- [ ] User Experience
- [ ] Test and Release
- [ ] Client Plugin
- [ ] Server Plugin
- [ ] Extensions
- [ ] Others

charlieblevins:
### WHY

In my use case, I have many different frpc binaries connecting to the same frps binary. Each frpc binary has a unique proxy name configured.

For this reason, it is useful to know which proxy (i.e. worker/control connection) is active. Without this PR, I can only know how many proxies are connected, but not which ones.

Fix #4986

### DIFF
--- a/pkg/metrics/prometheus/server.go
+++ b/pkg/metrics/prometheus/server.go
@@ -74,7 +74,7 @@ func newServerMetrics() *serverMetrics {
 			Namespace: namespace,
 			Subsystem: serverSubsystem,
 			Name:      "proxy_counts_detailed",
-			Help:      "The current proxy counts with proxy name label",
+			Help:      "The current number of proxies grouped by type and name",
 		}, []string{"type", "name"}),
 		connectionCount: prometheus.NewGaugeVec(prometheus.GaugeOpts{
 			Namespace: namespace,


pkg/metrics/prometheus/server.go
```diff
@@ -67,6 +70,12 @@ func newServerMetrics() *serverMetrics {
 			Name:      "proxy_counts",
 			Help:      "The current proxy counts",
 		}, []string{"type"}),
+		proxyCountDetailed: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Namespace: namespace,
+			Subsystem: serverSubsystem,
+			Name:      "proxy_counts_detailed",
```
charlieblevins, line 76: I used plural `proxy_counts_detailed` here for consistency with proxy_counts above

PR #5022, opened 2025-10-16
Fix CloseNotifyConn.Close function

zachwhaley:
### WHY

The `CloseNotifyConn.Close` function calls itself if the `closeFlag` is equal to `0` which would mean it would immediately swap the `closeFlag` value to `1` and never call `closeFn`.  It is unclear what the intent of this call to `cc.Close()` was but I assume it was meant to be a call to close the Conn object instead.

### DIFF
--- a/pkg/util/net/conn.go
+++ b/pkg/util/net/conn.go
@@ -149,7 +149,7 @@ func WrapCloseNotifyConn(c net.Conn, closeFn func()) net.Conn {
 func (cc *CloseNotifyConn) Close() (err error) {
 	pflag := atomic.SwapInt32(&cc.closeFlag, 1)
 	if pflag == 0 {
-		err = cc.Close()
+		err = cc.Conn.Close()
 		if cc.closeFn != nil {
 			cc.closeFn()
 		}


zachwhaley:
Sorry, I think I'm mistaken.  The recursive call would return and continue the original call to run closeFn.  But that still raises the question what is the purpose of the recursive call?

zachwhaley:
A more comprehensive fix to this could remove the recursive call and swap function like so:

```go
type CloseNotifyConn struct {
    net.Conn
    closeOnce sync.Once
    closeFn func()
}

func (cc *CloseNotifyConn) Close() (err error) {
    cc.closeOnce.Do(func() {
        err = cc.Conn.Close()
        if cc.closeFn != nil {
            cc.closeFn()
        }
    })
    return
}
```

Same goes for the `StatsConn.Close()` in this file

```go
type StatsConn struct {
    net.Conn

    closeOnce  sync.Once // changed from int64
}

func (statsConn *StatsConn) Close() (err error) {
    statsConn.closeOnce.Do(func() {
        err = statsConn.Conn.Close()
        if statsConn.statsFunc != nil {
            statsConn.statsFunc(statsConn.totalRead, statsConn.totalWrite)
        }
    })
    return
}
```

PR #5039, opened 2025-10-31
auth: add support for EntraID method

Saancreed:
### WHY

At LeanCode, we're using frp internally for exposing local development environments for external connections. There, we have a need to authenticate developers to cloud-hosted frps instance in a way that integrates with our existing authentication schemes while being as seamless for them as possible. To this end, we have implemented support for custom token-based authentication method that relies specifically on Microsoft Entra ID.

This method is similar in nature to existing OIDC method but has some additional benefits, namely:

* On the server side, we can easily restrict access by limiting it to users that can sign in to a specific Entra Application (representing frps) and optionally, only from specific Azure tenant.
* On the client side, we can retrieve the token credential using Azure Identity / MSAL, which allows us to authenticate using one of its many supported methods without having to pass any client credentials to frpc, most importantly Azure CLI.

Would there be any interest in accepting this, well, very vendor-specific piece of code? We think there's a chance others may also find some use for it.

### DIFF
--- a/pkg/auth/auth.go
+++ b/pkg/auth/auth.go
@@ -36,6 +36,8 @@ func NewAuthSetter(cfg v1.AuthClientConfig) (authProvider Setter, err error) {
 		if err != nil {
 			return nil, err
 		}
+	case v1.AuthMethodEntraID:
+		authProvider = NewEntraIDAuthSetter(cfg.AdditionalScopes, cfg.EntraID)
 	default:
 		return nil, fmt.Errorf("unsupported auth method: %s", cfg.Method)
 	}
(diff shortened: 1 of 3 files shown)


fatedier:
Thanks for the PR! We prefer not to add cloud-vendor–specific SDKs or features into the core. They make testing harder and create long-term maintenance burden that we don’t have the bandwidth to support. For this area, our direction is to provide extension points so provider-specific logic can live in external tools or plugins rather than in core.

Also, please open an issue to discuss the approach before submitting a PR. That helps us align on design and saves review time. Thanks!

Saancreed:
Thanks for the feedback. I understand that having vendor-specific code is generally not something fun to maintain. As an alternative, I've submitted a feature request with an idea that would still satisfy our needs while avoiding introducing such vendor-specific code: https://github.com/fatedier/frp/issues/5045

For now I'll close this PR, we can discuss this further in linked issue.

Issue #5045, opened 2025-11-03
[Feature Request] Auth method: CLI


Saancreed:
### Describe the feature request

On the client side, we have a need to be able to fetch PrivilegeKey from another (and often simple) program running on the same machine as the client. The idea would be to add a new, client-only auth method, let's call it `cli`, that runs an arbitrary program with given arguments, captures its standard output (stripping it of any trailing newlines), and uses it as the key. The credential retrieved this way could be used with either `token` or `oidc` auth method on the server side, depending on whether this program prints a pre-shared key or issues an access token.

For example:

```toml
# frps.toml
[auth]
method = "token"
token = "{{ .Envs.FRP_TOKEN }}"

# frpc.toml, retrieve credential from 1Password
[auth]
method = "cli"
cli.command = ["op", "item", "get", "frp", "--fields", "credential", "--reveal"]
```

or

```toml
# frps.toml
[auth]
method = "oidc"
oidc.issuer = "https://login.microsoftonline.com/{{ .Envs.FRP_AZURE_TENANT_ID }}/v2.0"
oidc.audience = "{{ .Envs.FRP_AZURE_APPLICATION_ID }}"

# frpc.toml, ask Azure CLI to issue access token
[auth]
method = "cli"
cli.command = ["az", "account", "get-access-token", "--tenant", "{{ .Envs.FRP_AZURE_TENANT_ID }}", "--scope", "{{ .Envs.FRP_AZURE_APPLICATION_ID }}/.default", "--query", "accessToken", "--output", "tsv"]
```

Optionally, we could consider supporting running the subprocess with additional environment variables but I wouldn't consider that a priority; on UNIX-like systems it's already easy enough to prepend something like `"env", "AZURE_CONFIG_DIR=/etc/azure"` to `command`.

For simplicity, we could assume that any non-zero exit code from the launched subprocess is an error, and that the output will already be properly (UTF-8?) encoded.

### Describe alternatives you've considered

* Explicit support for desired credential providers directly in frp
  * Rejected, frp shouldn't contain vendor-specific code in its core codebase.

* Some sort of plugin system, inspired by RPC-based server plugin system?
  * Complicated, hosting a full-blown HTTP server is something we'd like to avoid if simple CLI invocation does the job.

### Affected area

- [ ] Docs
- [ ] Installation
- [ ] Performance and Scalability
- [x] Security
- [ ] User Experience
- [ ] Test and Release
- [ ] Developer Infrastructure
- [ ] Client Plugin
[...]

PR #5050, opened 2025-11-10
Add `exec` value source type

Saancreed:
### WHY

https://github.com/fatedier/frp/issues/5045

Notably, the current version:

* uses simple `map[string]string` to represent additional environment variables
* also supports `dir` option that allows customizing the working directory of executed process (because adding it was pretty much free)
* does not consider `oidc` and `tokenSource` to be mutually exclusive; if both are present, the latter will take precedence
* uses `--allow-unsafe-token-source` CLI flag to enable usage of `exec` source type but the name isn't final… and maybe `file` should be considered unsafe too?
* requires the flag to be specified again by the caller when using `reload` command
* doesn't have any tests or documentation changes yet, I can add them once this is considered to be an acceptable approach

### DIFF
--- a/cmd/frpc/sub/root.go
+++ b/cmd/frpc/sub/root.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"slices"
 	"sync"
 	"syscall"
 	"time"
@@ -36,11 +37,18 @@ import (
 	"github.com/fatedier/frp/pkg/util/version"
 )
 
+type UnsafeFeature = string
+
+const (
+	TokenSourceExec UnsafeFeature = "TokenSourceExec"
+)
+
 var (
 	cfgFile          string
 	cfgDir           string
 	showVersion      bool
 	strictConfigMode bool
+	allowUnsafe      []UnsafeFeature
 )
 
 func init() {
@@ -48,6 +56,7 @@ func init() {
 	rootCmd.PersistentFlags().StringVarP(&cfgDir, "config_dir", "", "", "config directory, run one frpc service for each file in config directory")
 	rootCmd.PersistentFlags().BoolVarP(&showVersion, "version", "v", false, "version of frpc")
 	rootCmd.PersistentFlags().BoolVarP(&strictConfigMode, "strict_config", "", true, "strict config parsing mode, unknown fields will cause an errors")
+	rootCmd.PersistentFlags().StringSliceVarP(&allowUnsafe, "allow_unsafe", "", []string{}, "allowed unsafe features, one or more of: TokenSourceExec")
 }
 
 var rootCmd = &cobra.Command{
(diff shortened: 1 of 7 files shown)


fatedier:
My meaning is to use tokenSource as an internal configuration item for oidc.

cmd/frpc/sub/root.go
```diff
@@ -37,17 +37,19 @@ import (
...
+	strictConfigMode       bool
+	allowUnsafeTokenSource bool
 )
 
 func init() {
 	rootCmd.PersistentFlags().StringVarP(&cfgFile, "config", "c", "./frpc.ini", "config file of frpc")
 	rootCmd.PersistentFlags().StringVarP(&cfgDir, "config_dir", "", "", "config directory, run one frpc service for each file in config directory")
 	rootCmd.PersistentFlags().BoolVarP(&showVersion, "version", "v", false, "version of frpc")
 	rootCmd.PersistentFlags().BoolVarP(&strictConfigMode, "strict_config", "", true, "strict config parsing mode, unknown fields will cause an errors")
+	rootCmd.PersistentFlags().BoolVarP(&allowUnsafeTokenSource, "allow_unsafe_token_source", "", false, "enables usage of exec-based token source")
```
fatedier: Since we may add other unsafe capabilities in the future, I suggest generalizing the flag to `--allow-unsafe` and making it accept a list/array. For now we would only support "TokenSourceExec".

pkg/config/v1/value_source.go
```diff
@@ -91,3 +108,49 @@ func (f *FileSource) Resolve(_ context.Context) (string, error) {
...
+func (e *ExecSource) Resolve(ctx context.Context) (string, error) {
+	if err := e.Validate(); err != nil {
+		return "", err
+	}
+
+	cmd := exec.CommandContext(ctx, e.Command, e.Args...)
+	cmd.Dir = e.Dir
+
+	if len(e.Env) != 0 {
+		cmd.Env = os.Environ()
```
fatedier: It should not inherit the parent environment.

pkg/config/v1/value_source.go
```diff
@@ -27,13 +28,22 @@ import (
...
 // FileSource specifies how to load a value from a file.
 type FileSource struct {
 	Path string `json:"path"`
 }
 
+// ExecSource specifies how to get a value from another program launched as subprocess.
+type ExecSource struct {
+	Command string            `json:"command"`
+	Args    []string          `json:"args,omitempty"`
+	Dir     string            `json:"dir,omitempty"`
```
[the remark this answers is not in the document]
Saancreed: Sure, removed. I added it in the first place to support running things like `terraform output` but even that has `-chdir=DIR` global option that can be used for this purpose. I guess we can drop it until one finds a program that:

* is sensitive to current directory,
* doesn't allow changing it via CLI,
* could be useful for generating tokens for frp.

PR #5085, opened 2025-12-02
Print detailed information about connection termination to enable external programmes to track connection status from the log

kenvix:
### WHY

Currently, frps only prints detailed connection information upon connection establishment, whereas when  the connection is closed it merely displays `join connections closed`. 

This hinders external programmes such as fail2ban from tracking connection lifecycles effectively. This pull request will enable connection information to be printed upon termination as well, facilitating the tracking of connections.

### DIFF
--- a/server/proxy/proxy.go
+++ b/server/proxy/proxy.go
@@ -258,8 +258,12 @@ func (pxy *BaseProxy) handleUserTCPConnection(userConn net.Conn) {
 		})
 	}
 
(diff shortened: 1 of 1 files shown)


fatedier:
Relying on log format for external programs like fail2ban is fragile - log messages aren't a stable API and can change anytime.

kenvix:
I understand the risks, but using fail2ban to track logs generated by frps to against SSH brute-force attacks has become the most widely used solution.

This PR aims to improve the accuracy of detection of existing solutions.

PR #5175, opened 2026-02-12
auth/oidc: cache OIDC access token and refresh before expiry

shani1998:
### WHY --> [Feature Request](https://github.com/fatedier/frp/issues/5185)

When `auth.additionalScopes` includes `HeartBeats`, every heartbeat calls
`clientcredentials.Config.Token(ctx)` which creates a new `oauth2.ReuseTokenSource`
on each invocation — the token is never actually reused between calls. This causes
a redundant HTTP request to the OIDC token endpoint on every heartbeat interval.

### Changes

- Use `Config.TokenSource(ctx)` once at init to create a persistent `oauth2.TokenSource`
  that caches the token and only refreshes when it approaches expiry
- Wrap with `oauth2.ReuseTokenSourceWithExpiry` for a configurable early-refresh buffer
- Add `tokenRefreshAdvanceDuration` config option (default: 300s) to control how
  early before expiry the token is refreshed
- Add unit test with mock HTTP server verifying token is fetched only once across
  multiple login and ping calls

Client Config
```
[auth]
method = "oidc"
additionalScopes = ["HeartBeats"]

[auth.oidc]
clientID = "my-client"
clientSecret = "my-secret"
tokenEndpointURL = "https://idp.example.com/oauth/token"
# Optional: seconds before token expiry to refresh (default: 300)
# tokenRefreshAdvanceDuration = 300

[[proxies]]
name = "my-proxy"
type = "https"
...
```
## Test plan
- [x] Added `TestOidcAuthProviderCachesToken` with mock HTTP server proving only 1 token request is made across login + 5 pings
- [x] All existing auth tests pass (`go test ./pkg/auth/...`)
- [x] `go vet ./pkg/auth/...` clean
- [x] `go build` compiles cleanly
- [x] Manual testing with OIDC provider and `auth.additionalScopes = ["HeartBeats"]` — verified token is reused across heartbeats and refreshed only near expiry

### DIFF
--- a/pkg/auth/oidc_test.go
+++ b/pkg/auth/oidc_test.go
@@ -74,7 +74,7 @@ func TestOidcAuthProviderFallsBackWhenNoExpiry(t *testing.T) {
 	tokenServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 		requestCount.Add(1)
 		w.Header().Set("Content-Type", "application/json")
-		_ = json.NewEncoder(w).Encode(map[string]interface{}{ //nolint:gosec // test-only dummy token response
+		_ = json.NewEncoder(w).Encode(map[string]any{ //nolint:gosec // test-only dummy token response
 			"access_token": "fresh-test-token",
 			"token_type":   "Bearer",
 		})
@@ -98,7 +98,7 @@ func TestOidcAuthProviderFallsBackWhenNoExpiry(t *testing.T) {
 	r.NoError(err)
 	r.Equal("fresh-test-token", loginMsg.PrivilegeKey)
 
-	for i := 0; i < 3; i++ {
+	for range 3 {
 		pingMsg := &msg.Ping{}
 		err = provider.SetPing(pingMsg)
 		r.NoError(err)
@@ -116,7 +116,7 @@ func TestOidcAuthProviderCachesToken(t *testing.T) {
 	tokenServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 		requestCount.Add(1)
 		w.Header().Set("Content-Type", "application/json")
-		_ = json.NewEncoder(w).Encode(map[string]interface{}{ //nolint:gosec // test-only dummy token response
+		_ = json.NewEncoder(w).Encode(map[string]any{ //nolint:gosec // test-only dummy token response
 			"access_token": "cached-test-token",
 			"token_type":   "Bearer",
 			"expires_in":   3600,
@@ -145,7 +145,7 @@ func TestOidcAuthProviderCachesToken(t *testing.T) {
 	r.Equal(int32(1), requestCount.Load())
 
 	// Subsequent calls should also reuse the cached token
-	for i := 0; i < 5; i++ {
+	for range 5 {
 		pingMsg := &msg.Ping{}
 		err = provider.SetPing(pingMsg)
 		r.NoError(err)


shani1998:
> ### WHY --> [Feature Request](https://github.com/fatedier/frp/issues/5185)
> When `auth.additionalScopes` includes `HeartBeats`, every heartbeat calls `clientcredentials.Config.Token(ctx)` which creates a new `oauth2.ReuseTokenSource` on each invocation — the token is never actually reused between calls. This causes a redundant HTTP request to the OIDC token endpoint on every heartbeat interval.
> 
> ### Changes
> * Use `Config.TokenSource(ctx)` once at init to create a persistent `oauth2.TokenSource`
>   that caches the token and only refreshes when it approaches expiry
> * Wrap with `oauth2.ReuseTokenSourceWithExpiry` for a configurable early-refresh buffer
> * Add `tokenRefreshAdvanceDuration` config option (default: 300s) to control how
>   early before expiry the token is refreshed
> * Add unit test with mock HTTP server verifying token is fetched only once across
>   multiple login and ping calls
> 
> Client Config
> 
> ```
> [auth]
> method = "oidc"
> additionalScopes = ["HeartBeats"]
> 
> [auth.oidc]
> clientID = "my-client"
> clientSecret = "my-secret"
> tokenEndpointURL = "https://idp.example.com/oauth/token"
> 
> [[proxies]]
> name = "my-proxy"
> type = "https"
> ...
> ```
> 
> ## Test plan
> * [x]  Added `TestOidcAuthProviderCachesToken` with mock HTTP server proving only 1 token request is made across login + 5 pings
> * [x]  All existing auth tests pass (`go test ./pkg/auth/...`)
> * [x]  `go vet ./pkg/auth/...` clean
> * [x]  `go build` compiles cleanly
> * [x]  Manual testing with OIDC provider and `auth.additionalScopes = ["HeartBeats"]` — verified token is reused across heartbeats and refreshed only near expiry

blizard863:
Thank you for this PR. The caching optimization for the OIDC token source is a great improvement. However, I've identified a regression that affects OIDC providers that do not return an expires_in field in their token response.

In the golang.org/x/oauth2 library, when a token's Expiry is the zero value (which happens when expires_in is missing), it is treated as valid forever. This means tokenSource.Token() will continue to return the initial cached token indefinitely and will never attempt to refresh it. This is a critical issue because, even though the expires_in field is missing, the underlying JWT will still have its own exp claim. Once that time passes, the server-side OIDC verification will fail for operations like ping or establishing new work connections, as FRP will still be using the stale token.

The previous implementation, which called tokenGenerator.Token(ctx) on every request, did not have this dependency on expires_in and always fetched a fresh token, avoiding this problem.

To maintain compatibility with all OIDC providers, I suggest we add a fallback mechanism:
1. After creating the cached TokenSource, check if the initial token's Expiry is the zero value.
2. If it is, this indicates the provider did not return expires_in. In this case, we should fall back to the original behavior of not using the cached TokenSource and instead fetch a new token for each request.


This will ensure the optimization works for the majority of cases while preventing a regression for providers that follow the OAuth 2.0 spec by omitting the recommended expires_in field.

shani1998:
> Thank you for this PR. The caching optimization for the OIDC token source is a great improvement. However, I've identified a regression that affects OIDC providers that do not return an expires_in field in their token response.
> 
> In the golang.org/x/oauth2 library, when a token's Expiry is the zero value (which happens when expires_in is missing), it is treated as valid forever. This means tokenSource.Token() will continue to return the initial cached token indefinitely and will never attempt to refresh it. This is a critical issue because, even though the expires_in field is missing, the underlying JWT will still have its own exp claim. Once that time passes, the server-side OIDC verification will fail for operations like ping or establishing new work connections, as FRP will still be using the stale token.
> 
> The previous implementation, which called tokenGenerator.Token(ctx) on every request, did not have this dependency on expires_in and always fetched a fresh token, avoiding this problem.
> 
> To maintain compatibility with all OIDC providers, I suggest we add a fallback mechanism:
> 
> 1. After creating the cached TokenSource, check if the initial token's Expiry is the zero value.
> 2. If it is, this indicates the provider did not return expires_in. In this case, we should fall back to the original behavior of not using the cached TokenSource and instead fetch a new token for each request.
> 
> This will ensure the optimization works for the majority of cases while preventing a regression for providers that follow the OAuth 2.0 spec by omitting the recommended expires_in field.

Thanks for finding this out, added a fallback approach incase expiry_in set to zero.

PR #5215, opened 2026-03-07
refactor: fix modernize lint issues

alexandear:
### WHY

Encourage up-to-date Go idioms and cleaner code by adding the [`modernize`](https://golangci-lint.run/docs/linters/configuration/#modernize) linter.

I decided to disable the `omitzero` rule because fixing it leads to some behavioural changes. It can be fixed in a separate PR.

`modernize` issues were fixed via running the following command:

```console
$ golangci-lint run --enable-only modernize --fix
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -18,6 +18,7 @@ linters:
   - lll
   - makezero
   - misspell
+  - modernize
   - prealloc
   - predeclared
   - revive
@@ -47,6 +48,9 @@ linters:
       ignore-rules:
       - cancelled
       - marshalled
+    modernize:
+      disable:
+      - omitzero
     unparam:
       check-exported: false
   exclusions:
--- a/client/proxy/sudp.go
+++ b/client/proxy/sudp.go
@@ -32,7 +32,7 @@ import (
 )
 
 func init() {
-	RegisterProxyFactory(reflect.TypeOf(&v1.SUDPProxyConfig{}), NewSUDPProxy)
+	RegisterProxyFactory(reflect.TypeFor[*v1.SUDPProxyConfig](), NewSUDPProxy)
 }
 
 type SUDPProxy struct {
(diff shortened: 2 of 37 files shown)


fatedier:
Thanks for the contribution! Found a couple of issues:

### 1. `pkg/util/http/http.go` — `ParseBasicAuth` behavioral change

The manual `strings.Cut` simplification drops the `!ok` guard:

```go
// Before (original)
cs := string(c)
s := strings.IndexByte(cs, ':')
if s < 0 {
    return  // returns ("", "", false)
}
return cs[:s], cs[s+1:], true

// This PR
return strings.Cut(string(c), ":")
```

When the decoded string has no `:`, `strings.Cut` returns `(fullString, "", false)` — so `username` is no longer empty. Both callers (`pkg/util/vhost/http.go:275` and `pkg/util/tcpmux/httpconnect.go:65`) ignore the `ok` return value with `_`, which means a malformed `Proxy-Authorization` header could affect route matching.

The `golangci-lint --fix` auto-generated version is actually correct and preserves the guard:

```go
before, after, ok0 := strings.Cut(cs, ":")
if !ok0 {
    return
}
return before, after, true
```

This also matches the stdlib's own `parseBasicAuth` implementation in `net/http/request.go`.

### 2. `Makefile` — `make fix` target

`go fix -omitzero=false` is not a valid flag — `go fix` only accepts `-fix`. This causes `make fix` to always fail. Please don't add the `fix` target to the Makefile for now.

### 3. General request

Please don't mix manual modifications with auto-fix changes in the same commit. The linter auto-fix output should be committed as-is, and any manual optimizations (e.g. `maps.Clone` instead of `maps.Copy`, inlining `SplitSeq`, simplifying `WaitGroup.Go`) should be in a separate commit. This makes it much easier to review and verify correctness.

alexandear:
@fatedier thank you for the review. I fixed all the issues.

PR #5234, opened 2026-03-13
auth/oidc: fix eager token fetch at startup, add validation and e2e tests

fatedier:
## Summary
- Fix startup regression from #5175: `NewOidcAuthSetter` no longer eagerly fetches a token at construction time. Instead, an `oidcTokenSource` wrapper defers the first fetch to `SetLogin()` inside `loopLoginUntilSuccess()`, so transient IdP outages no longer prevent the client from starting.
- Add `ValidateOIDCClientCredentialsConfig` for early detection of misconfigured OIDC client credentials (missing clientID, invalid tokenEndpointURL, conflicting params).
- Add OIDC e2e tests with a self-contained mock OIDC server (discovery + JWKS + token endpoint, zero external dependencies):
  - Basic OIDC client credentials authentication
  - HeartBeats scope with authenticated pings
  - Adaptive fallback when provider omits `expires_in` (with token request count assertion)
  - Rejection of invalid client credentials

## Test plan
- [x] `go test ./pkg/auth/...` — unit tests pass (6 tests)
- [x] `go test ./pkg/config/v1/validation/...` — validation tests pass
- [x] `golangci-lint run` — 0 issues
- [x] `make e2e --ginkgo.focus="Feature: OIDC"` — 4 OIDC e2e tests pass
- [x] `make e2e` — full e2e suite passes (only pre-existing Ports Whitelist failure)


blizard863:
OIDC LGTM

PR #5240, opened 2026-03-16
fix: use ConstantTimeEqString for visitor auth key comparison

karesansui-u:
`server/visitor/visitor.go` compares the visitor auth key using `!=`,
while `pkg/auth/token.go` and `pkg/nathole/controller.go` use
`util.ConstantTimeEqString()` for the same `GetAuthKey` comparison.

This aligns the visitor auth path with the existing pattern.

### DIFF
--- a/server/visitor/visitor.go
+++ b/server/visitor/visitor.go
@@ -70,7 +70,7 @@ func (vm *Manager) NewConn(name string, conn net.Conn, timestamp int64, signKey 
 	defer vm.mu.RUnlock()
 
 	if l, ok := vm.listeners[name]; ok {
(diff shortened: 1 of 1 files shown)


karesansui-u:
The CI failure in `Ports Whitelist` test appears unrelated to this change — this PR only modifies the string comparison method in `server/visitor/visitor.go`, while the failing test is for port whitelisting in `test/e2e/v1/basic/server.go:89`.

The server log shows `tcp-port-unavailable: port not allowed` was correctly triggered, but the test received `nil` instead of an error, suggesting a race condition in the e2e test.

Could you re-run the CI? Happy to investigate further if needed.

PR #5261, opened 2026-03-25
nathole: return error when SetTTL fails in sendSidMessage

cuiweixie:
## Summary
- Propagate `SetTTL` errors in `sendSidMessage` instead of only logging and continuing, so UDP is not sent with an unintended TTL when TTL setup fails.
- Matches existing behavior for `TTL()` failures in the same path.

## Context
Per [Contributing](https://github.com/fatedier/frp/blob/dev/README.md#contributing), PR targets **dev**.

### DIFF
--- a/pkg/nathole/nathole.go
+++ b/pkg/nathole/nathole.go
@@ -372,6 +372,7 @@ func sendSidMessage(
 		err = uConn.SetTTL(ttl)
 		if err != nil {
 			xl.Tracef("set ttl error %v", err)
+			return err
 		} else {
 			defer func() {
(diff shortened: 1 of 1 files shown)


fatedier:
I don’t see a clear bug being fixed here. SetTTL failure in this path is currently handled as a best-effort degradation, and this PR changes it into a hard error without a concrete failure scenario or evidence that it is the right tradeoff.

In general, please avoid sending PRs of this kind unless there is a clear abnormal case, reproducible issue, or demonstrated benefit. Otherwise they add review and maintenance cost without solving a concrete problem.

PR #5283, opened 2026-04-09
server: add proxy protocol support for bind and vhost HTTP/HTTPS listeners

simplysoft:
Add vhostHTTPSProxyProtocol and vhostHTTPProxyProtocol config options that enable PROXY protocol (v1/v2) parsing on the bind and vhost listeners.

When enabled, the real client IP is extracted from the PROXY protocol header sent by an upstream load balancer and used as the source address for proxied connections. The IP flows through the existing StartWorkConn.SrcAddr mechanism to frpc, making it available in X-Forwarded-For headers (https2http/https2https plugins) and via transport.proxyProtocolVersion for raw TCP proxies.

Uses the existing github.com/pires/go-proxyproto dependency (already used by frpc for sending proxy protocol to local services).


simplysoft:
comment to keep PR open

PR #5301, opened 2026-04-29
plugin/client: add mcp_stdio plugin

ailenshen:
### WHY

Most MCP (Model Context Protocol) servers in the wild ship as stdio
binaries — `npx -y @org/mcp-...`, `uvx ...` and similar — and only
know how to speak JSON-RPC over stdin/stdout. To make them reachable
by remote MCP clients like Claude Desktop or the Anthropic API today,
you have to run a separate stdio↔HTTP translator (e.g. `mcp-proxy`)
locally alongside frpc.

This PR adds an `mcp_stdio` client plugin that folds that translation
into frpc itself, so a single frpc process is enough to expose any
number of stdio MCP servers as remote Streamable HTTP endpoints — no
extra translator process and no new runtime dependencies.

### What it does

- Spawns the configured stdio command lazily on the first request.
- For each inbound HTTP POST, writes the body as a JSON-RPC line to
  the child's stdin and returns the next stdout line as the HTTP
  response. Notifications (frames without an `id`) are accepted with
  HTTP 202 and produce no stdio output.
- Optional `idleTimeoutSeconds`: kill the child after N seconds of
  inactivity. The MCP `initialize` handshake is cached and replayed
  when the child is respawned, so sessions stay healthy across reaps
  and the next spawn picks up any updated package version on its own
  (useful with `@latest` style npx commands).

### Configuration

\`\`\`toml
[[proxies]]
name = "apple-notes"
type = "tcp"
remotePort = 3101
[proxies.plugin]
type = "mcp_stdio"
command = ["npx", "-y", "@modelcontextprotocol/server-everything"]
idleTimeoutSeconds = 300
\`\`\`

### Tested with

- Real apple-notes and filesystem MCP servers end-to-end through a
  public frps, reachable from Claude Desktop and claude.ai.
- Unit tests cover JSON-RPC classification, dispatch, notifications,
  and the respawn-with-replay path.
- \`golangci-lint run ./pkg/plugin/client/... ./pkg/config/v1/...\`
  passes.


Qodo-Free-For-OSS:
Hi, `dispatch()` performs blocking stdin/stdout I/O while holding `p.mu` and has no timeouts/cancellation, so a stalled child can hang one request indefinitely and block all subsequent requests; it can also prevent `Close()` from acquiring the mutex to kill the child during shutdown.

**Severity:** action required | **Category:** reliability

**How to fix:** Add timeouts and cancellation

**Agent prompt to fix** - you can give this to your LLM of choice:

> ### Issue description
> `dispatch()` holds a global mutex while doing unbounded blocking reads from the child’s stdout. If the child never produces a newline, requests hang indefinitely and all subsequent requests to the plugin block behind the mutex; `Close()` can also block waiting for the same mutex.
> 
> ### Issue Context
> The HTTP handler currently does not use `r.Context()`, and the child process is started without a context/deadline. `http.Server` only sets `ReadHeaderTimeout`, so body reads and backend stdio waits can last forever.
> 
> ### Fix Focus Areas
> - pkg/plugin/client/mcp_stdio.go[111-181]
> - pkg/plugin/client/mcp_stdio.go[203-255]
> - pkg/plugin/client/mcp_stdio.go[313-321]
> 
> ### What to change
> - Stop holding `p.mu` across blocking I/O (e.g., implement a single worker goroutine that owns the child pipes and processes requests via a channel, or lock only to swap/check child state and do I/O outside the lock).
> - Plumb request cancellation into dispatch (e.g., accept a `context.Context` from `handle()` and use it to bound waiting for stdout).
> - Add read/write deadlines for the child I/O and/or a per-request timeout to fail fast and respawn/kill the child on timeout.
> - Ensure `Close()` can always terminate quickly (e.g., kill the child before waiting on anything that could require acquiring `p.mu`, or make the worker goroutine observe a close signal).

We noticed a couple of other issues in this PR as well - happy to share if helpful.

---
*Found by [Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source) code review. FYI, Qodo is free for open-source.*

ailenshen:
Thanks for the review. The concerns are valid — the original implementation held the global mutex across blocking I/O, which could stall all requests if the child process hung, and prevented `Close()` from acquiring the lock during shutdown.

Addressed in the latest commit by refactoring to a worker goroutine pattern:
- A single `worker()` goroutine exclusively owns the child process and all its I/O — no mutexes needed
- `dispatch()` communicates with the worker via channels and respects `r.Context()` for cancellation throughout
- A 30-second per-request read timeout kills and respawns the child on a stalled response (the goroutine spawned for the read never leaks since the buffered channel absorbs its result after the pipe closes)
- `Close()` simply signals `closeCh` and waits on `workerDone` — no lock contention possible
- Idle reaping is merged into the worker's `select` loop, removing the separate `reapLoop` goroutine

ailenshen:
Thanks for the detailed review. All four warnings addressed in the latest commit:

**1. Body size limit** — wrapped `r.Body` with `http.MaxBytesReader(1 MiB)` before `io.ReadAll`.

**2. Replay handshake blocking** — extracted a `readLine` helper closure (goroutine + buffered channel + `time.After` select) and replaced the bare `childOut.ReadBytes` in the `ensureChild` replay path.

**3. Pretty-printed JSON splitting frames** — added `json.Compact` in `handle()` after reading the body; multi-line input is collapsed to a single line before being forwarded to the child's stdin.

**4. Response not matched by id** — added a `readResponse` helper that loops, discarding lines whose JSON-RPC `id` does not match the request's `id`, until the correct response arrives. The MCP spec ([transports](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports)) explicitly permits servers to send notifications (e.g. `notifications/progress`) before the response to a request, so reading only the next line was a protocol-level bug. A single `time.After` deadline is shared across iterations so chatty servers cannot extend the timeout.

fatedier:
Thanks for the contribution.

We are not planning to include this kind of application-layer integration in frp core for now. We may consider supporting similar use cases through a more suitable plugin mechanism in the future.

ailenshen:
Thanks for the clarification. Totally understood — application-layer integrations like this are better kept outside of frp core. I'll maintain it in my own fork. Appreciate the review and the thoughtful feedback from the community along the way.

PR #5305, opened 2026-05-09
fix: reset UDP read deadline on write activity to prevent idle timeout

silverpoetry:
## Problem

The `Forwarder` function in `pkg/proto/udp/udp.go` has a 30-second `SetReadDeadline` in `writerFn` that is only reset when data is read from the local service. When traffic is unidirectional (the remote side sends packets through the proxy, but the local service has no data to send back), the read deadline expires after 30 seconds and the UDP socket is destroyed — even though the write direction is actively processing packets.

This affects real-world scenarios like game streaming audio (Sunshine/Moonlight) where ACK packets arrive but audio may be silent for 30+ seconds.

## Fix

After a successful `Write` to the local UDP socket, reset the read deadline. This ensures the socket stays alive as long as there is activity in **either** direction.

## Testing

Reproduced and verified with a Python test script using the FRP tunnel locally. Stock v0.68.1 drops the socket after 29 seconds of unidirectional traffic; the fix keeps it alive.


fatedier (approved):
Looks good to me. The fix correctly refreshes the UDP read deadline on successful write activity, so active unidirectional traffic is not treated as idle.

Issue #5315, opened 2026-05-17
[Feature Request] Add a new 'xudp' mode for optimized P2P hole punching


fscarmen:
### Describe the feature request

Please consider adding a brand-new xudp mode inspired by Hysteria Realms:

* https://v2.hysteria.network/docs/advanced/Realms/

* Lightweight Signaling: Peers use a shared token/name to exchange addresses via the server, which exits the flow entirely once the UDP hole is punched.

* Multi-STUN Prediction: Allows configuring multiple STUN servers to observe mapping rules, significantly improving P2P success rates under Symmetric (predictable) NATs.

* Decoupled Lifecycle: The established P2P tunnel runs independently without being affected by frpc-frps control channel disconnections.

fabricnp:
Hi maintainers,

I've implemented an `xudp` proxy mode addressing this feature request. The implementation lives in my fork: https://github.com/fabricnp/frp/pull/1

**What it does:**
- **Multi-STUN port prediction** ??? queries multiple STUN servers serially from the same local port to detect predictable Symmetric NAT port allocation patterns, then augments hole-punch candidates with predicted ports
- **Realm rendezvous** ??? reuses the existing `NatHoleController` as a lightweight signaling relay (shared token exchange), then cleanly exits the data flow
- **Decoupled QUIC tunnel** ??? the punched tunnel uses `context.Background()` in an isolated goroutine, surviving control channel drops (ALPN: `frp-xudp`)

**Integration approach:**
- Hooks into the existing `MakeHole` concurrency pattern (unbuffered channels, select timeouts)
- Standard proxy/visitor factory registration (`ProxyTypeXUDP`, `VisitorTypeXUDP`)
- No new wire protocol messages ??? reuses all existing `NatHole*` message types
- Falls back gracefully to standard hole-punch behavior when prediction fails

**Quality:**
- Passes `go vet`, `gofmt`, and all existing unit tests
- 16 new table-driven tests for port prediction logic
- Full documentation at `doc/xudp-mode.md`
- Addressed all automated review findings (response demux safety, lifecycle ownership, goroutine leak prevention)

Requesting feedback from maintainers. Happy to adjust the approach based on your architectural preferences.

PR #5337, opened 2026-05-27
feat: add xudp proxy mode with multi-STUN prediction and decoupled tunnel

fabricnp:
This adds a new `xudp` proxy type that targets predictable Symmetric NATs (ref #5315).

The core idea: query multiple STUN servers serially from the same local port, observe the port allocation pattern, and use that to predict where the next hole-punch packet should land. The prediction is advisory and the system falls back to standard MakeHole behavior if no consistent pattern is detected.

The tunnel itself runs on QUIC (ALPN `frp-xudp`) in a goroutine with its own background context, so it stays alive even if the control channel to frps drops temporarily.

**How it works:**

1. frps acts as rendezvous (reuses NatHoleController, no new messages on the wire)
2. Both sides do STUN discovery, exchange addresses through frps, then frps exits the data path
3. If multiple STUN servers are configured, the client observes port deltas and appends predicted ports to the candidate list before punching
4. After the hole is established, a QUIC session is set up with a decoupled lifecycle

**Files added:**

- `pkg/nathole/xudp.go` - multi-STUN discovery, port prediction, hole punch wrapper
- `pkg/nathole/xudp_test.go` - unit tests for the prediction logic
- `client/proxy/xudp.go` - client proxy implementation
- `client/visitor/xudp.go` - visitor with tunnel session management
- `server/proxy/xudp.go` - server proxy (same pattern as xtcp)
- `doc/xudp-mode.md` - docs explaining the architecture and config

Registers as `ProxyTypeXUDP` / `VisitorTypeXUDP` through the standard factory system. No changes to existing proxy types or the wire protocol.

Tested with `go vet`, `gofmt`, and the existing test suite plus 16 new tests for the port prediction edge cases.

### DIFF
--- a/client/proxy/xudp.go
+++ b/client/proxy/xudp.go
@@ -1,0 +1,224 @@
+// Copyright 2024 The frp Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build !frps
+
+package proxy
+
+import (
+	"context"
+	"net"
+	"reflect"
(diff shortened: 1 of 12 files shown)


fabricnp:
Closing this PR. After further review, the multi-STUN prediction approach adds complexity that outweighs the benefit for the current use case. Going to rethink the rendezvous design without the prediction overhead. Appreciate the review feedback.

PR #5371, opened 2026-06-19
fix(client): honor no_proxy when connecting to the server

### LINKED ISSUE #4964: [Feature Request] 不支持 no_proxy 环境变量

发现 frpc 会读取环境变量的 http_proxy ，但是 no_proxy 列表中的 ip 和 domain 不会 ignore 掉。

_No response_

### Affected area

- [ ] Docs
- [ ] Installation
- [ ] Security
- [ ] User Experience
- [ ] Test and Release
- [ ] Client Plugin
- [ ] Server Plugin
- [ ] Extensions
- [ ] Others

Rilomilo:
## What

frpc reads the `http_proxy` environment variable to decide whether to dial frps through a proxy (`pkg/config/v1/client.go`: `ProxyURL = EmptyOr(ProxyURL, os.Getenv("http_proxy"))`), but it never consulted `no_proxy`. So a server that should be reached directly — `localhost`/loopback, or a host explicitly listed in `no_proxy` — was still routed through the proxy, which usually fails.

This evaluates the `no_proxy` rules against the server address before applying the proxy. When the address is excluded, frpc connects directly.

- Uses `golang.org/x/net/http/httpproxy` (already an existing dependency) for the matching, so domains, suffixes, CIDRs and the `*` wildcard all behave like every other tool that honors `http_proxy`/`no_proxy`.
- `httpproxy` also implicitly exempts `localhost` and loopback addresses, so connecting to a loopback frps no longer goes through the proxy even when `no_proxy` doesn't list it.

## Why

Closes #4964. It is also surprising in general: honoring `http_proxy` while ignoring `no_proxy` breaks the de-facto convention shared by curl, the Go standard library, etc.

## Behavior change

Only affects setups where `http_proxy`/`transport.proxyURL` is set. For an excluded destination the connection is now direct instead of proxied. Setups without a proxy configured are unaffected.

## Test

- Added a unit test (`client/connector_test.go`) covering: no proxy configured, default proxying, localhost/loopback exemption, and `no_proxy` matching by IP, CIDR, domain and `*`.
- Manually verified end-to-end: with `http_proxy=http://proxy:8080` and `no_proxy=127.0.0.1` set, a stock frpc connecting to a loopback frps fails (the control connection is sent to the proxy), while the patched frpc connects and tunnels successfully.

```
$ go test ./client/ -run TestProxyURLForServer
ok      github.com/fatedier/frp/client
```

`golangci-lint` is clean.

### DIFF
--- a/client/connector.go
+++ b/client/connector.go
@@ -18,6 +18,8 @@ import (
 	"context"
 	"crypto/tls"
 	"net"
+	"net/url"
+	"os"
 	"strconv"
 	"strings"
 	"sync"
@@ -27,6 +29,7 @@ import (
 	fmux "github.com/hashicorp/yamux"
 	quic "github.com/quic-go/quic-go"
 	"github.com/samber/lo"
+	"golang.org/x/net/http/httpproxy"
 
 	v1 "github.com/fatedier/frp/pkg/config/v1"
 	"github.com/fatedier/frp/pkg/msg"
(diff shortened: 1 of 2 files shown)


fatedier:
I don't think we should further extend the current `http_proxy` environment-variable behavior.

In frpc, this proxy setting only applies to the frpc-to-frps connection, while frpc may also connect to local services or other endpoints. Using a process-wide environment variable for this makes the scope ambiguous.

I prefer making this explicit via `transport.proxyURL`. We may also consider weakening or deprecating the automatic fallback from `http_proxy` in the future.

Rilomilo:
That's a fair point — a process-wide environment variable is an ambiguous scope for what is really just the frpc-to-frps connection, and steering users toward an explicit `transport.proxyURL` (and eventually phasing out the `http_proxy` fallback) is a cleaner direction than extending the env-var behavior. Closing this out. Thanks for the thoughtful review!

PR #5393, opened 2026-07-03
fix: close doneCh for a control replaced before it starts (#5391)

### LINKED ISSUE #5391: frps: permanent goroutine + memory leak when a client re-logins after its connection died silently — login handler blocks forever in oldCtl.WaitClosed() (tcpMux)
## Bug Description

When a client's control connection dies **silently** (no FIN/RST reaches the server — e.g. NAT mapping expiry, DPI middlebox drop, pulled cable) and frpc then reconnects with the same `runID`, frps permanently leaks one goroutine plus the entire old `Control` object graph (yamux session, TLS state, bufio buffers — ~30–50 KB) **on every login attempt**, and the client can never reconnect again until frps is restarted.

### Mechanism

1. On re-login with the same `runID`, `RegisterControl` calls `ctlManager.Add()`, which calls `oldCtl.Replaced(newCtl)`, and then blocks in `oldCtl.WaitClosed()` (`server/service.go:789`).
2. `Control.Replaced()` closes only `ctl.sessionCtx.Conn` — with `tcpMux` enabled that is the **yamux stream, not the underlying TCP connection** (`server/control.go:194-199`). Closing a yamux stream locally half-closes it; the old control's `msgDispatcher` read on that stream never returns on a silently-dead path, so `worker()` never reaches cleanup and `doneCh` is never closed. The login handler goroutine is now blocked forever.
3. `ctl.Start()` and the `LoginResp` write happen **after** `RegisterControl` returns (`server/service.go:483-496`). So the reconnecting client never receives `LoginResp`, times out, and retries ~30s later.
4. The retry login blocks in `WaitClosed()` on the *previous* control — whose `Start()` **was never called** (its handler is itself stuck in step 3), so its `doneCh` can literally never close. From this point the chain can never unwind, even after every TCP connection involved has long been dead and closed by the kernel. Each retry adds one permanently-stuck goroutine + one retained `Control` graph.

The result is a self-sustaining loop: a single transient network fault permanently wedges that client's `runID` on the server, the client hammers re-logins forever, and frps memory grows linearly without bound. File descriptors do **not** grow (the sockets do get closed eventually) — only heap and goroutines.

In our production deployment (frps v0.69.1, ~14 IoT devices behind consumer ISPs): frps grew from 10 MiB to **760 MiB** in days; one log day contained **6,817 `replaced by client`** entries against **8 `client exit success`** entries. Devices caught in the loop lose their tunnels entirely until frps is restarted.

`goroutine` pprof profile (count exactly equals the number of `replaced by client` events; taken after all TCP connections were dead):

```
152 @ 0x101051660 0x100fe8648 0x100fe81e4 0x1016e2210 0x1016e2205 0x1016dff68 0x1010597d4
#	github.com/fatedier/frp/server.(*Control).WaitClosed		server/control.go:285
#	github.com/fatedier/frp/server.(*Service).RegisterControl	server/service.go:789
#	github.com/fatedier/frp/server.(*Service).handleConnection	server/service.go:466
```

### Notes

- Reproduces with **default configuration** (single client, no TLS force, no auth token, no plugins).
- Server-side bug: reproduced with both frpc v0.65.0 and v0.69.1 clients.
- Does **not** reproduce with `transport.tcpMux = false` on both sides: `Replaced()` then closes the real TCP socket, the old control's reader unblocks immediately, cleanup completes (`client exit success` for every replace), goroutine count and memory stay flat.
- Code is unchanged in the current `dev` branch, so the bug is present there as well.

### Possible fix directions

- Make `Replaced()` tear down the old control's underlying mux session (or set a read deadline on its stream) so cleanup can always complete, and/or
- Don't block the login path indefinitely: bound the `oldCtl.WaitClosed()` wait, and/or
- Guarantee `doneCh` is eventually closed even for controls whose `Start()` was never called (e.g. close it from the error path of `handleConnection`).

## frpc Version

0.69.1 (also reproduced with 0.65.0)

## frps Version

0.69.1

## System Architecture

darwin/arm64 (repro), linux/amd64 (production)

## Configurations

frps.toml:

```toml
bindAddr = "127.0.0.1"
bindPort = 17000
webServer.addr = "127.0.0.1"
webServer.port = 17500
webServer.user = "admin"
webServer.password = "admin"
webServer.pprofEnable = true
log.to = "console"
```

frpc.toml (points at the relay below, which forwards to frps):

```toml
serverAddr = "127.0.0.1"
serverPort = 17001
user = "testclient"
[[proxies]]
name = "svc"
type = "tcp"
localIP = "127.0.0.1"
localPort = 65000
remotePort = 20001
```

## Logs

```
19:55:20.330 [I] [server/service.go:758] [8400eb1b9892f71f] client login info: ip [127.0.0.1:61410] version [0.69.1] ...
19:55:23.331 [I] [server/service.go:758] [8400eb1b9892f71f] client login info: ip [127.0.0.1:61416] version [0.69.1] ...
19:55:23.331 [I] [server/control.go:196] [8400eb1b9892f71f] replaced by client [8400eb1b9892f71f]
19:55:28.433 [I] [server/service.go:758] [8400eb1b9892f71f] client login info: ip [127.0.0.1:61422] version [0.69.1] ...
19:55:28.433 [I] [server/control.go:196] [8400eb1b9892f71f] replaced by client [8400eb1b9892f71f]
19:55:35.819 [I] [server/control.go:196] [8400eb1b9892f71f] replaced by client [8400eb1b9892f71f]
...
```

`client exit success` almost never appears (in a 2.5-minute run: 9 `replaced by client`, 1 `client exit success`, 8 goroutines permanently parked in `WaitClosed`).

## Steps to reproduce

1. Save `relay.py` — a TCP relay that simulates a silently-dying network path (DPI/NAT behaviour): every 3 s it closes the frpc side of the connection but **abandons the frps side without closing it**, so frps never receives FIN/RST:

```python
import asyncio, sys

UPSTREAM = ('127.0.0.1', 17000)   # real frps
RESET_INTERVAL = 3.0
abandoned = []                     # keep server-side sockets open forever

async def pump(reader, writer):
    try:
        while True:
            data = await reader.read(65536)
            if not data:
                break
            writer.write(data)
            await writer.drain()
    except Exception:
        pass

async def handle(reader, writer):
    up_reader, up_writer = await asyncio.open_connection(*UPSTREAM)
    t1 = asyncio.create_task(pump(reader, up_writer))
    t2 = asyncio.create_task(pump(up_reader, writer))
    await asyncio.sleep(RESET_INTERVAL)
    t1.cancel(); t2.cancel()
    writer.close()                 # frpc side: sees EOF, reconnects
    abandoned.append((up_reader, up_writer))  # frps side: silently abandoned

async def main():
    server = await asyncio.start_server(handle, '127.0.0.1', 17001)
    await server.serve_forever()

asyncio.run(main())
```

2. `frps -c frps.toml`
3. `python3 relay.py`
4. `frpc -c frpc.toml`
5. Wait 2–3 minutes. frps logs show repeated `client login info` + `replaced by client` with (almost) no `client exit success`; frpc never manages to reconnect (no `LoginResp` ever arrives).
6. Check the goroutine profile — the number of goroutines parked in `Control.WaitClosed` equals the number of `replaced by client` events and never decreases, even after stopping the relay and frpc entirely:

```
curl -su admin:admin "http://127.0.0.1:17500/debug/pprof/goroutine?debug=1" | grep -A3 RegisterControl
```

## Affected area

- [x] Others

maximilize:
### WHY

Fixes #5391.

When a control connection dies silently with `tcpMux` (no FIN/RST reaches frps) and the client re-logins with the same `runID`, frps permanently parks a goroutine in `Control.WaitClosed` and retains the old `Control` graph — the client can never reconnect until frps restarts.

**Root cause** (on `dev`, `7fe152e`). `RegisterControl` blocks in `oldCtl.WaitClosed()` (`server/service.go:788`) via `ControlManager.Add` → `old.Replaced()` (`server/control.go:56-64`, `:194-199`). `WaitClosed` (`control.go:285`) returns only when `doneCh` closes, and `doneCh` is closed **only** in `worker()` (`control.go:337`), which runs **only** from `Start()` (`control.go:186`). `Start()` is called after `RegisterControl` returns (`service.go:496`) — so a control that gets `Replaced()` while its own login handler is still blocked in `RegisterControl` (i.e. it never reached `Start()`) can never have its `doneCh` closed by anything. Each re-login then parks one more goroutine on the previous, never-started control, forever (matches the reporter's pprof: every parked goroutine sits in `WaitClosed` ← `RegisterControl`).

**Fix.** Track whether a control was ever started, and when a never-started control is `Replaced()`, close its `doneCh` there (a never-started control has no worker/dispatcher goroutines, so nothing else will). `Start()` now also skips a control that was already replaced. `doneCh` is closed through a small once-guarded helper to avoid a double close. This lets the replacement chain unwind so re-logins succeed instead of wedging.

**Verification** (built on this branch, `--network none`, using the reporter's `relay.py` that abandons the frps side of each connection):

- New unit test `TestControlReplacedBeforeStartUnblocksWaitClosed` — `WaitClosed` returns after `Replaced` on a never-started control. Fails on `dev` (blocks 2s → timeout), passes with this change. `TestControlReplacedAfterStartLeavesDoneToWorker` guards that a started control's `doneCh` is still owned by `worker`.
- End-to-end, tcpMux on, 4-minute run + 150s after stopping the relay/client:
  - **before:** goroutines parked in `WaitClosed` stay put permanently (still parked 150s after every connection is dead), `client exit success` freezes after the first cleanup, the client stops reconnecting.
  - **after:** parked goroutines drain to 0, cleanup resumes (`exit success` tracks re-logins), the client keeps reconnecting.
- `go test ./server/`, `go vet ./server/`, `gofmt` all clean.

**Scope note.** This fixes the permanent leak (a control replaced before it starts). For a control that *was* started and whose reader is stuck on a silently-dead tcpMux stream, cleanup still relies on yamux keepalive to close the session (~30s), after which its `doneCh` closes normally. If you want that path hardened too, tearing down the mux session in `Replaced()` (or bounding the login-path `WaitClosed()`) would remove that ~30s window — glad to fold it in.

### DIFF
--- a/server/control.go
+++ b/server/control.go
@@ -227,6 +227,9 @@ func (ctl *Control) Replaced(newCtl *Control) {
 	xl := ctl.xl
 	xl.Infof("replaced by client [%s]", newCtl.runID)
 	ctl.runID = ""
+	// Closing a yamux stream only half-closes it and does not wake a blocked local read.
+	// Expire the read first so the control worker can always finish its cleanup.
+	_ = ctl.sessionCtx.Conn.SetReadDeadline(time.Now())
 	ctl.sessionCtx.Conn.Close()
 
 	// If this control was never started, its worker will never run and would
--- a/server/control.go
+++ b/server/control.go
@@ -66,13 +66,17 @@ func (cm *ControlManager) Add(runID string, ctl *Control) (old *Control) {
 	return
 }
 
-// we should make sure if it's the same control to prevent delete a new one
-func (cm *ControlManager) Del(runID string, ctl *Control) {
+// Del removes ctl only if it is still the control registered for runID, to avoid
+// deleting a newer one. It reports whether it actually removed ctl, so callers can
+// tell whether ctl was still the active control for this run id.
+func (cm *ControlManager) Del(runID string, ctl *Control) (deleted bool) {
 	cm.mu.Lock()
 	defer cm.mu.Unlock()
 	if c, ok := cm.ctlsByRunID[runID]; ok && c == ctl {
 		delete(cm.ctlsByRunID, runID)
+		return true
 	}
+	return false
 }
 
 func (cm *ControlManager) GetByID(runID string) (ctl *Control, ok bool) {
--- a/server/control.go
+++ b/server/control.go
@@ -149,9 +149,26 @@ type Control struct {
 
 	mu sync.RWMutex
 
+	// started is true once Start has launched the worker goroutine.
+	// replaced is true once this control has been superseded by a newer login.
+	// doneClosed guards doneCh against a double close.
+	started    bool
+	replaced   bool
+	doneClosed bool
+
 	xl     *xlog.Logger
 	ctx    context.Context
 	doneCh chan struct{}
+}
+
+// closeDone closes doneCh exactly once, unblocking WaitClosed.
+func (ctl *Control) closeDone() {
+	ctl.mu.Lock()
+	defer ctl.mu.Unlock()
+	if !ctl.doneClosed {
+		ctl.doneClosed = true
+		close(ctl.doneCh)
+	}
 }
 
 func NewControl(ctx context.Context, sessionCtx *SessionContext) (*Control, error) {
@@ -177,6 +194,17 @@ func NewControl(ctx context.Context, sessionCtx *SessionContext) (*Control, erro
 
 // Start starts the control session workers after login succeeds.
 func (ctl *Control) Start() {
+	ctl.mu.Lock()
+	if ctl.replaced {
+		// This control was already replaced by a newer login before it started.
+		// worker would never observe a live connection, and Replaced has already
+		// closed doneCh, so there is nothing to run.
+		ctl.mu.Unlock()
+		return
+	}
+	ctl.started = true
+	ctl.mu.Unlock()
+
 	go func() {
 		for i := 0; i < ctl.poolCount; i++ {
 			// ignore error here, that means that this control is closed
@@ -196,6 +224,21 @@ func (ctl *Control) Replaced(newCtl *Control) {
 	xl.Infof("replaced by client [%s]", newCtl.runID)
 	ctl.runID = ""
 	ctl.sessionCtx.Conn.Close()
+
+	// If this control was never started, its worker will never run and would
+	// never close doneCh. That happens when its own login handler is still
+	// blocked in RegisterControl waiting on a previous control (e.g. after a
+	// control connection died silently with tcpMux, so the old reader never
+	// unblocks). Close doneCh here so WaitClosed returns and the replacement
+	// chain can unwind, instead of leaking a goroutine plus the whole Control
+	// graph permanently. See #5391.
+	ctl.mu.Lock()
+	ctl.replaced = true
+	notStarted := !ctl.started
+	ctl.mu.Unlock()
+	if notStarted {
+		ctl.closeDone()
+	}
 }
 
 func (ctl *Control) RegisterWorkConn(conn *proxy.WorkConn) error {
@@ -334,7 +377,7 @@ func (ctl *Control) worker() {
 	metrics.Server.CloseClient()
 	ctl.sessionCtx.ClientRegistry.MarkOfflineByRunID(ctl.runID)
 	xl.Infof("client exit success")
-	close(ctl.doneCh)
+	ctl.closeDone()
 }
 
 func (ctl *Control) registerMsgHandlers() {
--- a/server/control_test.go
+++ b/server/control_test.go
@@ -15,12 +15,19 @@
 package server
 
 import (
+	"context"
 	"net"
+	"os"
+	"sync"
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/require"
+
+	v1 "github.com/fatedier/frp/pkg/config/v1"
 	"github.com/fatedier/frp/pkg/msg"
 	"github.com/fatedier/frp/pkg/util/xlog"
+	"github.com/fatedier/frp/server/registry"
 )
 
 // newTestControl builds a minimal Control sufficient for lifecycle tests.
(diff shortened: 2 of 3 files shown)


maximilize:
Good catch — confirmed. With the `doneCh` fix, a control replaced before it ever started now unblocks its login handler; if that handler then fails to write the login response on the connection `Replaced` closed, the error path called `MarkOfflineByRunID(m.RunID)` unconditionally. Since the registry keys on run id (not the control instance), that would offline the run id now owned by the replacement control.

Fixed in 12b0728: `ControlManager.Del` now reports whether it removed the *still-registered* control, and `MarkOfflineByRunID` is gated on that, so a superseded handler can't offline its replacement. Added `TestControlManagerDelOnlyRemovesActiveControl`. (This was latent for any rapid re-login where a control is superseded while still waiting on an older one; the `doneCh` fix just makes that recovery path reachable.)

barkure:
Hi @maximilize — I tested a small incremental hardening on top of this PR and opened maximilize/frp#1.

It sets an expired read deadline before closing a replaced control stream, so an already-started control blocked in a yamux read reaches the existing worker cleanup immediately. This complements this PR's fix for controls replaced before `Start()`.

I also ran the reporter's `relay.py` with `tcpMux` enabled and a 3-second reset interval for 60 seconds. frpc repeatedly logged in with the same run ID; all 12 pprof samples reported zero goroutines in `Control.WaitClosed`. Five seconds after stopping frpc and the relay, both `Control.WaitClosed` and `Service.RegisterControl` were still zero.

Validation: `make test`, `go test -race ./server -count=1`, and `git diff --check` all pass. Would you be open to merging maximilize/frp#1 into this branch?

maximilize:
Reviewed and folded into this PR, @barkure — kept as your commit, so your authorship is preserved.

`SetReadDeadline(time.Now())` before `Conn.Close()` in `Replaced()` is the right complement: closing a yamux stream half-closes it and doesn't wake a blocked local read, so an already-started control's worker would otherwise wait for keepalive before `doneCh` closes. Expiring the read deadline wakes it immediately — together with the pre-`Start()` path both cases are covered now. Nice independent repro with the pprof samples.

maximilize:
Superseded by #5424 (merged), which reworks the control replacement lifecycle and closes #5391 directly. That change covers the same paths this PR touched — including a half-open relay e2e harness matching the reporter's `relay.py` scenario — so this branch is now redundant and conflicts against `dev`. Closing it.

Thanks @fatedier for the comprehensive fix, and @barkure for the already-started hardening (`SetReadDeadline` before close) and the pprof repro that were folded in here.

Issue #4531, opened 2024-11-12
UDP over UDP tunnel


scriptchildie:
### Describe the feature request

Hi fantastic project. 

I was wondering if there is a way to tunnel UDP traffic over UDP instead of the TCP Port.

Ideally want to have another option in the config file bindUdpPort

Please let me know if it's already implemented and I missed it. 

Thank you  

### Affected area

- [ ] Docs
- [ ] Installation
- [X] Performance and Scalability
- [ ] Security
- [ ] User Experience
- [ ] Test and Release
- [ ] Developer Infrastructure
- [ ] Client Plugin
- [ ] Server Plugin
- [ ] Extensions
- [ ] Others

scriptchildie:
nice one thanks!

CalunVier:
> https://github.com/fatedier/frp/blob/dev/conf/frpc_full_example.toml#L83
> 
> kcp and quic are based on UDP.

Can a purely UDP connection be provided? Both KCP and QUIC have packet loss retransmission mechanisms, but packet loss can be considered a virtue. In some cases, upper-layer applications may already have their own packet loss retransmission mechanisms, which could introduce additional overhead.
能否提供纯粹的UDP连接？kcp 和 quic都有丢包重传机制，但丢包是一种美德（某些情况下，上层应用会包含丢包重传机制，这会带来额外的负担？）

Issue #4174, opened 2024-04-20
wireguard over frp,performance is lower than wireguard over rathole.


linnuxx:
### Bug Description

What could be the matter?
wgserver mtu 1300.
wgclient mtu 1300.
frps/frpc are installed on debian12.
go version go1.22.2 linux/amd64.

### frpc Version

0.57.0

### frps Version

0.57.0

### System Architecture

linux/amd64

### Configurations

# frps.toml
bindPort = 7000

# frpc.toml
serverAddr = "ipvps"
serverPort = 7000

[[proxies]]
name = "wg"
type = "udp"
localIP = "192.168.1.237"
localPort = 51820
remotePort = 51821
transport.useCompression = true

### Steps to reproduce

1. 
2. 
3. 
...

### Affected area

- [ ] Docs
- [ ] Installation
- [X] Performance and Scalability
- [ ] Security
- [ ] User Experience
- [ ] Test and Release
- [ ] Developer Infrastructure
- [ ] Client Plugin
- [ ] Server Plugin
- [ ] Extensions
- [X] Others

fatedier:
Thank you for your testing. It would be more helpful if you could provide complete data and various tests of controlled variables. Otherwise, we won't be able to do anything.

fatedier:
I am not familiar with wireguard and rathole, so I cannot provide any additional advice.

linnuxx:
> I am not familiar with wireguard and rathole, so I cannot provide any additional advice.

you are familiar with frp, but for some reason the speed of working with frp is lower.

linnuxx:
> Thank you for your testing. It would be more helpful if you could provide complete data and various tests of controlled variables. Otherwise, we won't be able to do anything.
![ScreenShot_20240429173814](https://github.com/fatedier/frp/assets/154101918/dcf56ba0-4a53-4e87-b811-3031d448047b)

ratholeserver:
# server.toml
[server]
bind_addr = "0.0.0.0:8388" # `2333` specifies the port that rathole listens for clients

[server.services.my_nas_ssh]
type = "udp"
token = "key" # Token that is used to authenticate the client for the service. Change to an arbitrary value.
bind_addr = "0.0.0.0:51821" # `5202` specifies the port that exposes `my_nas_ssh` to the Internet
nodelay = true

ratholeclient:
# client.toml
[client]
remote_addr = "46.174.53.140:8388" # The address of the server. The port must be the same with the port in `server.bind_addr`

[...]

PR #5401, opened 2026-07-09
udp proxy: relay packets as unreliable QUIC datagrams (RFC 9221)

zt-chen:
### Problem

The `udp` proxy wraps every UDP packet in a `msg.UDPPacket` and writes it over a **single reliable work-connection stream** (`server/proxy/udp.go`, `client/proxy/udp.go`). When the tunneled protocol already tolerates loss — VPN DTLS (OpenConnect/ocserv, AnyConnect), WireGuard, RTP/QUIC media — this adds a reliability layer it neither needs nor wants:

- **Head-of-line blocking**: one lost segment stalls every subsequent packet for all users of that proxy.
- **Redundant retransmission**: the outer stream re-sends payloads the inner protocol has already given up on.

On a lossy, high-RTT path the single stream is bounded by the Mathis limit (`~MSS / (RTT·√loss)`), so throughput collapses far below the raw path capacity. This is the root cause behind #4531 (request to carry UDP unreliably) and #4174 (WireGuard-over-frp slower than the alternative); switching the transport to `kcp`/`quic` doesn't help because those only make the *underlay* UDP — the proxied payload still rides a reliable stream.

#### Why

A UDP proxy should preserve UDP's delivery semantics end to end. Today it doesn't: it turns an unreliable datagram service into a reliable, ordered byte stream, which is a poor fit for the very protocols people most often tunnel over `type = udp` (VPN DTLS, WireGuard, RTP, QUIC media). Those protocols run their own loss recovery, so frp's reliability is pure overhead — and worse than neutral, because a single reliable stream converts packet loss into latency and head-of-line stalls that cascade across every flow sharing the proxy.

The alternatives don't address it:

- **Transport tuning (`kcp`/`quic`, window/keepalive knobs)** changes only the underlay; the payload is still framed onto one reliable stream, so the HoL blocking and redundant retransmission remain.
- **A dedicated stream per user flow** would remove cross-flow HoL but still imposes per-flow reliability/ordering the inner protocol doesn't want, plus stream-setup cost per UDP "session."
- **A second raw UDP side-channel** (à la the KCP/FakeTCP requests) means new ports, new firewall/NAT holes, and separate auth — exactly what frp's existing multiplexed connection avoids.

QUIC datagrams (RFC 9221) fit because the QUIC transport frp already supports carries them **on the same authenticated, congestion-controlled connection**, with no extra ports and no new handshake — while giving each packet unreliable, unordered delivery. It's the minimal change that makes a UDP proxy behave like a UDP proxy, and it's opt-in so nothing changes for users who want today's behavior.

### Change

Adds an opt-in `quicDatagrams` option to `udp` proxies. When enabled and the frpc↔frps connection is QUIC, packets are relayed as **unreliable QUIC datagrams (RFC 9221)** instead of over the work-connection stream, using a compact binary frame (proxy name + user address + raw payload — no JSON, no base64). Lost packets stay lost; there is no outer ordering or retransmission.

```toml
# frpc.toml
transport.protocol = "quic"

[[proxies]]
name = "vpn-dtls"
type = "udp"
localIP = "127.0.0.1"
localPort = 10443
remotePort = 10443
quicDatagrams = true
```

### Design / compatibility

- **Negotiated per work connection.** The client advertises `NewProxy.udp_datagram`; the server confirms with `StartWorkConn.udp_datagram` only when *that* work connection actually supports datagrams. Either side that lacks support (older peer, non-QUIC transport) simply keeps the existing stream path — no config error, no breakage. Both message fields are additive JSON, so mixed-version deployments interoperate.
- **Per-packet fallback.** A frame that exceeds a conservative datagram budget (1180 bytes) is sent over the stream instead, so oversized packets are never dropped. The budget deliberately sits below the minimum QUIC packet size: quic-go's `SendDatagram` size check uses the current MTU estimate *without* subtracting packet/AEAD overhead, so frames in a narrow band just under the MTU are accepted and then silently dropped at packing time — the fixed budget avoids that blackhole.
- **Interop with stream wrappers.** Datagram mode is disabled automatically when `useEncryption` or `useCompression` is set (those are stream transforms), falling back to the stream.
- Requires `transport.protocol = "quic"`; no effect for tcp/kcp/websocket transports.

### Tests

- New `pkg/proto/udp/datagram_test.go`: frame round-trip (incl. empty/binary payloads, IPv6), encode errors, malformed-frame rejection.
- Functional (local frps+frpc, UDP echo): all payload sizes 64–1450 round-trip correctly; the stream path is unaffected; datagram mode is confirmed on both ends.
- Load (loopback): 60 Mbit/s sustained at 0% loss through the datagram lane.
- Real-world A/B on a lossy ~150 ms intercontinental path, same binaries, offered 55 Mbit/s: **datagram lane 50.7 Mbit/s @ 7.8% loss vs. the stream lane 25.7 Mbit/s @ 51% loss** — roughly 2× goodput and near the raw path capacity, with materially lower jitter.

### Notes

Feature is fully opt-in and off by default; existing `udp` proxies are unchanged. Happy to adjust the config field name, the datagram budget constant, or split the message/config additions into a separate commit if that's preferred.

### DIFF
--- a/client/connector.go
+++ b/client/connector.go
@@ -130,6 +130,7 @@ func (c *defaultConnectorImpl) Open() error {
 				MaxIdleTimeout:     time.Duration(c.cfg.Transport.QUIC.MaxIdleTimeout) * time.Second,
 				MaxIncomingStreams: int64(c.cfg.Transport.QUIC.MaxIncomingStreams),
 				KeepAlivePeriod:    time.Duration(c.cfg.Transport.QUIC.KeepalivePeriod) * time.Second,
+				EnableDatagrams:    true, // for udp proxies with quicDatagrams (RFC 9221)
 			})
 		if err != nil {
 			return err
--- a/client/proxy/udp.go
+++ b/client/proxy/udp.go
@@ -45,7 +45,9 @@ type UDPProxy struct {
 	// include msg.UDPPacket and msg.Ping
 	sendCh   chan msg.Message
 	workConn net.Conn
-	closed   bool
+	// non-nil while relaying via unreliable QUIC datagrams (RFC 9221)
+	dgMux  *udp.DatagramMux
+	closed bool
 }
 
 func NewUDPProxy(baseProxy *BaseProxy, cfg v1.ProxyConfigurer) Proxy {
@@ -73,6 +75,10 @@ func (pxy *UDPProxy) Close() {
 
 	if !pxy.closed {
 		pxy.closed = true
+		if pxy.dgMux != nil {
+			pxy.dgMux.Unregister(pxy.cfg.Name)
+			pxy.dgMux = nil
+		}
 		if pxy.workConn != nil {
 			pxy.workConn.Close()
 		}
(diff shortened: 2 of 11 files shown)


zt-chen:
Pop to prevent stale, 
Maintainer, will this get merged?

zt-chen:
pop to prevent stale

zt-chen:
pop to prevent stale

zt-chen:
pop to prevent stale

PR #5415, opened 2026-07-11
web: restore Vite client declarations for both dashboards

barkure:
## Summary

- restore the Vite client declarations to the TypeScript programs of both dashboards
- move `web/frpc/env.d.ts` and `web/frps/env.d.ts` into `src/` so they are covered by the existing `src/**` include patterns
- drop the `declare module '*.vue'` wildcard shim

## Root cause

The 2024 bundler-mode migration changed both dashboard tsconfig include lists to `src/**`, leaving the root `env.d.ts` files outside the TypeScript programs. Moving the declarations under `src/` restores the `vite/client` types for both dashboards without any tsconfig change, and keeps frpc and frps consistent with exactly one declaration entry each.

The wildcard `*.vue` module shim is not needed by the current `vue-tsc` toolchain and would only mask real SFC prop types under plain `tsc`, so it is removed.

## Verification

Rebased onto the latest `dev` and ran a clean verification (Node v22.23.1, npm 10.9.8):

```sh
npm ci
npm run type-check --workspace=frpc
npm run build --workspace=frpc
npm run lint --workspace=frpc
npm run type-check --workspace=frps
npm run build --workspace=frps
npm run lint --workspace=frps
```

All commands passed for both workspaces.


fatedier:
Thanks for the contribution. I reviewed this PR against the current `dev` branch and traced the dashboard TypeScript configuration history.

The declaration visibility cleanup is useful, but please revise the implementation before merge.

In clean, isolated clones, both current `origin/dev` (`84b907a198d143be26bc5869d6046e82f32de2ea`) and this PR head (`2497ce519461eea4319b7b237a152f7de50e2038`) passed `npm ci`, the frpc type-check, build, and lint commands. A direct type probe also confirmed that the current `vue-tsc` toolchain preserves real SFC props and emits without a wildcard declaration. The proposed `declare module '*.vue'` only acts as a broad fallback for plain `tsc`, where specific component prop types are lost.

Historically, the root `env.d.ts` files were explicitly included. The 2024 bundler-mode migration changed both dashboard include lists to `src/**` but left the root declaration files behind. The clean, consistent fix is to move the declarations into the existing `src` include for both dashboards.

Please make these changes:

1. For frpc, keep `web/frpc/src/env.d.ts` but reduce it to only:

```ts
/// <reference types="vite/client" />
```

Then delete the old `web/frpc/env.d.ts` root file.

2. Apply the same move to frps:

```text
web/frps/env.d.ts -> web/frps/src/env.d.ts
```

The new frps file should also contain only the Vite client reference.

3. Remove the `declare module '*.vue'` wildcard shim. No `tsconfig` include change is needed because both declarations will be under the existing `src/**` include patterns. Naming the files `src/vite-env.d.ts` instead is also acceptable, as long as each dashboard has exactly one declaration entry and both use the same convention.

4. Update the PR title/body to describe restoring the Vite client declarations to the TypeScript programs for both dashboards. The current `vue-tsc` TS2307 claim is not reproducible with the current lockfile; if it should remain in the description, please provide the exact Node/npm environment and clean failure output.

5. Rebase onto the latest `dev` and rerun clean verification for both workspaces:

```sh
npm ci
npm run type-check --workspace=frpc
npm run build --workspace=frpc
npm run lint --workspace=frpc
npm run type-check --workspace=frps
npm run build --workspace=frps
npm run lint --workspace=frps
```

This leaves one active Vite declaration per dashboard, keeps frpc and frps consistent, and avoids a broad Vue module fallback that the current `vue-tsc` workflow does not need.

barkure:
Thanks for the detailed review. All points addressed:

1. `web/frpc/env.d.ts` moved to `web/frpc/src/env.d.ts`, reduced to just `/// <reference types="vite/client" />`.
2. Same move applied to frps: `web/frps/env.d.ts` → `web/frps/src/env.d.ts`, same content.
3. The `declare module '*.vue'` wildcard shim is removed; no tsconfig change needed since both files now sit under the existing `src/**` include.
4. PR title and body updated to describe restoring the Vite client declarations for both dashboards; the unreproducible TS2307 claim is dropped.
5. Rebased onto the latest `dev` and reran the clean verification for both workspaces (`npm ci`, `type-check`, `build`, `lint`) with Node v22.23.1 / npm 10.9.8 — all passed.

The final diff is just the two file moves with unchanged content.

PR #5455, opened 2026-07-28
ci: attach provenance and SBOM attestations to the published images

kobihikri:
Hi, and thanks for frp.

`.github/workflows/build-and-push-image.yml` pushes four images on every release — `fatedier/frpc` and `fatedier/frps` to Docker Hub, and the same two to `ghcr.io` — but the manifests carry no provenance or SBOM attestation. Someone pulling `fatedier/frps:latest` has no way to check that the image was built by this workflow, from this repository, at that tag.

frp is usually deployed as an edge service with a public listener, and the image is very often pulled straight into a `docker run` or compose file. That combination is why this seemed worth raising rather than left alone.

The change is two lines on each of the two build steps:

```yaml
      - name: Build and push frpc
        uses: docker/build-push-action@v7
        with:
          ...
          push: true
          provenance: mode=max
          sbom: true
```

I used BuildKit's own attestation support rather than a separate signing action, for a specific reason: you push the same build to two registries in one step, and BuildKit attaches the attestation to the image manifest itself, so both Docker Hub and GHCR get it without any extra plumbing. It also means **no permissions change** — the workflow keeps `contents: read` exactly as it is, and nothing needs `id-token`.

After a release, consumers can check with:

```
docker buildx imagetools inspect ghcr.io/fatedier/frps:<tag> --format '{{ json .Provenance }}'
```

Two honest caveats:

- `mode=max` records the full build, including the Dockerfile and build arguments. That is normally what you want for a public image, but if any of your build args were ever sensitive it is worth a look before merging — `provenance: true` gives a smaller record if you would prefer.
- Attestations add an extra manifest to the index. Docker Hub and GHCR both handle this, but a very old registry mirror in front of them might not. Worth knowing if you have one in the path.

I have not claimed a SLSA level here — the attestation is what BuildKit produces, and how the overall build should be characterised is your call.

Disclosure: I used AI assistance to help spot this and prepare the change, and I read the workflow and its push targets myself.

### DIFF
--- a/.github/workflows/build-and-push-image.yml
+++ b/.github/workflows/build-and-push-image.yml
@@ -67,6 +67,8 @@ jobs:
           file: ./dockerfiles/Dockerfile-for-frpc
           platforms: linux/amd64,linux/arm/v7,linux/arm64,linux/ppc64le
           push: true
+          provenance: mode=max
+          sbom: true
           tags: |
             ${{ env.TAG_FRPC }}
             ${{ env.TAG_FRPC_GPR }}
@@ -78,6 +80,8 @@ jobs:
           file: ./dockerfiles/Dockerfile-for-frps
           platforms: linux/amd64,linux/arm/v7,linux/arm64,linux/ppc64le
           push: true
+          provenance: mode=max
+          sbom: true
           tags: |
             ${{ env.TAG_FRPS }}
             ${{ env.TAG_FRPS_GPR }}


kobihikri:
**Correction — I got a fact wrong in this PR, and I would rather flag it myself than let it sit.**

I wrote that the pushed manifest "carries no provenance or SBOM attestation". That is half wrong, and the wrong half matters.

**Provenance is already there.** For *public* repositories, `docker/build-push-action` adds provenance attestations with `mode=max` by default — [Docker's documentation](https://docs.docker.com/build/ci/github-actions/attestations/) states it plainly: *"Public repos: provenance attestations with `mode=max` are automatically added"*. I checked published images and they do already carry attestation manifests. So the `provenance: mode=max` line in my diff makes existing behaviour **explicit**; it does not add anything new.

**The SBOM is genuinely new.** That part stands — the same page says *"SBOM attestations aren't automatically added to the image"*, and `sbom: true` is what enables them.

I also wrote in the caveats that `provenance: true` gives "a smaller record". That is wrong as well: `true` resolves to max on a public repo, and the smaller setting is `provenance: mode=min`.

**So the honest description of this PR is:** it adds an SBOM attestation, and pins the provenance mode explicitly instead of relying on the default. Both are still defensible — an explicit line means the behaviour will not change quietly if the default ever does — but it is a smaller change than my description implied, and you should judge it on that basis rather than on what I originally wrote.

Happy to retitle and rewrite the description accordingly, or to close this if the SBOM alone is not worth the diff to you. Either is fine — just say which and I will act on it.

Apologies for the inaccuracy. It was caught by a maintainer reviewing the same change on another project, and they were right to.

fatedier:
Thanks for the clarification. I think the cleanest version of this PR would be to keep `sbom: true` and remove the explicit `provenance: mode=max` lines.

For a public repository, `docker/build-push-action` already generates max-level provenance by default, so those lines only restate the existing behavior. The SBOM is the actual new capability introduced by this PR.

Could you please update the diff and the PR title/description accordingly?

PR #5502, opened 2026-08-20
fix(health): reset failed count after successful check

racequite:
### WHY

`Monitor.failedTimes` is incremented after every failed health check, but it is never reset after a successful check.

With `MaxFailed = 3`, the following sequence is handled incorrectly:

1. health check fails
2. health check fails
3. health check succeeds
4. health check fails

The fourth check can immediately mark the backend as failed because `failedTimes` is still `2`. This makes `MaxFailed` behave like a cumulative failure threshold since monitor creation, instead of a threshold for consecutive failures.

This is related to #5434, which was closed without being merged.

### DIFF
--- a/client/health/health.go
+++ b/client/health/health.go
@@ -119,6 +119,7 @@ func (monitor *Monitor) checkWorker() {
 
 		if err == nil {
 			xl.Tracef("do one health check success")
+			monitor.failedTimes = 0
 			if !monitor.statusOK && monitor.statusNormalFn != nil {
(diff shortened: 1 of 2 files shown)


racequite:
@fatedier Hi, could you take a look at this small health-check fix?

failedTimes currently survives a successful health check, so MaxFailed can behave as a cumulative failure threshold instead of requiring consecutive failures.

fatedier (approved):

PR #5513, opened 2026-08-28
feat: enhance websocket support (multiple paths and native wss)

aries-shen:
- frps: support multiple websocket upgrade paths via transport.websocketPaths,
  and frpc: support a custom path via transport.websocketPath.
- frps: support wss (websocket over TLS) natively. After TLS termination frps
  inspects the connection for a websocket upgrade request (matching
  transport.websocketPaths) and hands it to the websocket listener, so wss
  clients can connect directly without a separate TLS-terminating proxy.
- Add IsWebsocketRequest matcher, TryWebsocket peeker that preserves initial
  bytes, and WebsocketListener.HandleConn to upgrade an already-TLS-terminated
  connection.

新增/增强 websocket 支持（多路径与原生 wss）

- frps 端通过 transport.websocketPaths 支持多个 websocket 升级路径，frpc 端通过
  transport.websocketPath 支持自定义路径。
- frps 原生支持 wss（基于 TLS 的 websocket）：TLS 终止后，frps 检测连接是否为
  websocket 升级请求（匹配 transport.websocketPaths）并交给 websocket 监听器处理，
  使 wss 客户端无需额外的 TLS 终止反向代理即可直接连接。
- 新增 IsWebsocketRequest 匹配器、可保留初始字节的 TryWebsocket 预读器，以及用于
  已终止 TLS 连接的 WebsocketListener.HandleConn。

### DIFF
--- a/client/connector.go
+++ b/client/connector.go
@@ -215,7 +215,7 @@ func (c *defaultConnectorImpl) realConnect() (net.Conn, error) {
 	switch protocol {
 	case "websocket":
 		protocol = "tcp"
-		dialOptions = append(dialOptions, libnet.WithAfterHook(libnet.AfterHook{Hook: netpkg.DialHookWebsocket(protocol, "")}))
+		dialOptions = append(dialOptions, libnet.WithAfterHook(libnet.AfterHook{Hook: netpkg.DialHookWebsocket(protocol, "", c.cfg.Transport.WebsocketPath)}))
 		dialOptions = append(dialOptions, libnet.WithAfterHook(libnet.AfterHook{
 			Hook: netpkg.DialHookCustomTLSHeadByte(tlsConfig != nil, lo.FromPtr(c.cfg.Transport.TLS.DisableCustomTLSFirstByte)),
 		}))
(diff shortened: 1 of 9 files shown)


aries-shen:
Addressed the two review warnings from the automated check:

1. **Per-connection `TryWebsocket` peek (perf).** The wss detection on the TLS listener now only runs when `transport.websocketPaths` is explicitly configured. Previously `ServerTransportConfig.Complete()` defaulted `websocketPaths` to `[/~!frp]`, so every TLS connection (TLS is on by default) went through the peek in the accept loop. I removed that default; the plaintext websocket listener still serves the default path `/~!frp` via the muxer, while the TLS-port peek (`server/service.go`) is now gated behind `len(cfg.Transport.WebsocketPaths) > 0`. Default frps therefore pays no per-connection peeking cost, and the 3s probe only applies when the operator opts into wss.

2. **Client `websocketPath` not normalized.** Added leading-`/` normalization in `ClientTransportConfig.Complete()` (mirroring the server side), so a value like `custom/path` is normalized to `/custom/path` and produces a valid URL in `DialHookWebsocket`.

PR #5527, opened 2026-09-09
protocol: derive v2 control channel keys from X25519 key agreement

shani1998:
### WHY

The v2 control channel derives its AEAD keys with

```go
hkdf.Key(sha256.New, key, transcriptHash, info, AEADKeySize)   // pkg/util/net/conn.go
```

`key` is `auth.EncryptionKey()`, which returns `[]byte(resolved.Token)`. `resolved.Token`
is only populated inside `if resolved.Method == v1.AuthMethodToken` (`pkg/auth/auth.go`),
so with `auth.method = oidc` it is empty on both sides.

The salt is `HashCryptoTranscript(clientHelloPayload, serverHelloPayload)` over two frames
sent in cleartext before encryption starts, and `info` is a constant. For OIDC deployments
every input to the KDF is then public, so an observer of the handshake derives the same
directional keys. This matters most behind TLS-terminating infrastructure, which sees the
hellos.

### WHAT

Add an ephemeral X25519 key share to `ClientHello`/`ServerHello` and prepend the ECDH
output to the HKDF input keying material. Token deployments keep contributing their
pre-shared secret and additionally gain forward secrecy.

Both `newControlReadWriter`s refuse to open a v2 channel when the resulting keying
material is empty, so a stripped key share fails closed instead of silently falling back
to a publicly derivable key.

### Compatibility

With `auth.method = token`, wire-compatible in both directions: a peer that offers no key
share is answered without one, the token still keys the channel, and the transcripts match
because both sides hash the raw hello payloads.

With an auth method that has no pre-shared secret, a peer carrying this change will refuse
to talk to a v2 peer that predates it, rather than negotiating a keyless channel. That is
the combination the change exists to fix, so it fails closed deliberately.

### Limitations

The exchange is unauthenticated on its own. An active attacker who rewrites both hellos can
substitute their own key shares and man-in-the-middle the channel, exactly as before this
change — the empty-IKM check makes the stripping variant fail closed, it does not make the
exchange authenticated. This removes passive key recovery, not active attack. Binding the
exchange to the authenticated identity needs a transcript confirmation after login and is
left out to keep this change small.

Out of scope, both still keyed by the raw auth token: the v1 control channel
(`NewCryptoReadWriter`) and work connections using `transport.useEncryption`.

### Tests

`pkg/proto/wire`: shared secret agreement with an empty auth key, per-session uniqueness,
the no-key-share fallback, rejection of an unsolicited server key share, a stripped client
key share leaving no keying material, malformed key-share lengths on both sides, and
low-order points.

`test/e2e/v1/basic/wire.go`: v2 without a token, v2 with a token, v1 with a token.

`go test ./pkg/proto/wire/... ./pkg/util/net/... ./client/... ./server/...` and
`ginkgo --focus="WireProtocol|BinaryUDPPacket|ControlReplacement" ./test/e2e` pass.

### DIFF
--- a/client/control_session.go
+++ b/client/control_session.go
@@ -16,6 +16,7 @@ package client
 
 import (
 	"context"
+	"crypto/ecdh"
 	"errors"
 	"fmt"
 	"io"
@@ -138,6 +139,7 @@ func (d *controlSessionDialer) exchangeLogin(conn net.Conn, loginMsg *msg.Login)
 	var wireConn *wire.Conn
 	var clientHello wire.ClientHello
 	var clientHelloPayload []byte
+	var clientKeyPriv *ecdh.PrivateKey
 
 	if d.common.Transport.WireProtocol == wire.ProtocolV2 {
 		if err := wire.WriteMagic(conn); err != nil {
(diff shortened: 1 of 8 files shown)


shani1998:
Opened #5532 with the problem description and the design questions, per the contributing guidelines. Happy to adjust the approach based on that discussion before this is reviewed.

